PR #24, opened 2015-05-15
Add check for non-empty text to fix runtime error

fedorov:
This issue caused runtime error while reconstructing prostate phantom dataset.

On a related note: after fixing this issue, the operation times out before getting the reconstructed volume back from the server. Should the timeout parameter be configurable? Not sure if it has any negative consequences, since even after the timeout the reconstructed volume is received in Slicer when ready.

### DIFF
--- a/PlusRemote/PlusRemote.py
+++ b/PlusRemote/PlusRemote.py
@@ -973,7 +973,8 @@ class PlusRemoteWidget(ScriptedLoadableModuleWidget):
     if command.GetResponseMessage():
       statusText = statusText + command.GetResponseMessage()
     else:
-      statusText = statusText + command.GetResponseText()
+      if command.GetResponseText():
+        statusText = statusText + command.GetResponseText()
     self.replyBox.setPlainText(statusText)
 
   def onGetCaptureDeviceCommandResponseReceived(self, command, q):


lassoan:
Timeout unlocks buttons so that you can press them again. How long is the reconstruction typically?

Have you tried live reconstruction? (if you know the region of interest then the slices are pasted immediately and you get the reconstructed volume in seconds)

fedorov:
On Fri, May 15, 2015 at 4:54 PM, Andras Lasso <notifications@github.com>
wrote:

> Timeout unlocks buttons so that you can press them again. How long is the
> reconstruction typically?
>

I didn't time it, more than a minute. Right now we use Plus 32bit build,
since we need trakSTAR drivers. I am actually not sure why reconstruction
works at all with 32 bit - the dataset is quite large, about 500M. It will
probably be better to control data collection from Slicer GUI vie 32-bit
PlusLauncher, but run a 64-bit reconstructor from command line (or from a
custom Slicer module).

> Have you tried live reconstruction? (if you know the region of interest
> then the slices are pasted immediately and you get the reconstructed volume
> in seconds)
>

No, we haven't tried that. For now, it is just proof of concept. We have
acquired trakSTAR and did all the setup/calibration for the first patient
study that we plan for Jul 9 (just to get patient data with freehand
tracked TRUS). Depending on how that goes, we will look into optimizing the
workflow.

PR #48, opened 2015-10-01
re 42: Added line and distance to the BreachWarning module.

brudfors:


lassoan:
Thank you. This requires some more work, because there should not be any redundant information stored in the BreachWarning node (for example, the ruler already stores it s color, thickness, etc.), a node should not create another node (BreachWarning should not create the ruler node, but the logic should create a ruler node if show line is requested on the GUI), the BreachWarning node cannot "store" the ruler node but it should just have a node reference to it, etc. I've already started on cleaning this up. Hopefully I can finish it by tomorrow.

brudfors:
I suspected that there would be some flaws with my implementation. The 
redundancy that you mention, for example, is obvious - sorry. But I 
didn't know that a node shouldn't create another node (now I do).

I will have a thorough look at your modifications!

Mikael

PR #78, opened 2016-05-10
Changes to MarkupsToModelNode

aniqah:
In ReadXMLAttributes function, there is a sequence of if statements that
are mututally exclusive. This change adds else statements to ensure that
no unnecessary conditional checks are made.


lassoan:
1. It's safer to create a separate std::stringstream object in each if/else branch to make sure no data can be accidentally left in a stream in case of unexpected file contents.

Something like this should work:
```
else if ( ! strcmp( attName, "Something" ) )  
{
  std::stringstream nameString(attValue);
  nameString >> this->Something;  
}  
```

2. For serializing predefined types (such as model type), define string values. In the XML we should not have modelType="1" but something like this instead: modelType="surface". See for example:
https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx#L183..L194

PR #83, opened 2016-06-06
Fixes #81: Allowing pivot and spin calibration in arbitrary order.

### LINKED ISSUE #81: Spin calibration button not active
Currently the spin calibration becomes active only when pivot calibration is done with the same input selections. But when we want to calibrate a stylus using recorded sequences, we want to change the input to the sequence that contains spin calibration data. But at this point the spin calibration button becomes inactive. I think the spin calibration button should always be active, and just suggest users to do pivot calibration first.

mholden8:


PivotCalibration/Logic/vtkSlicerPivotCalibrationLogic.cxx
```diff
@@ -42,6 +42,13 @@
 
 
 static const double PARALLEL_ANGLE_THRESHOLD_DEGREES = 20.0;
+// Note: If the needle orientation protocol changes, only the definitions of shaftAxis and secondaryAxes need to be changed
+// Define the shaft axis and the secondary shaft axis
+// Current needle orientation protocol dictates: shaft axis -z, orthogonal axis +x
+// If StylusX is parallel to ShaftAxis then: shaft axis -z, orthogonal axis +y
+static const double SHAFT_AXIS[ 3 ] = { 0, 0, -1 };
+static const double ORTHOGONAL_AXIS[ 3 ] = { 1, 0, 0 };
+static const double BACKUP_AXIS[ 3 ] = { 1, 0, 0 };
```
lassoan: Is that correct that BACKUP_AXIS is the same as ORTHOGONAL_AXIS?

mholden8: Good catch. It should be { 0, 1, 0 }, as described in the comment above it. I'll update the pull request.

PivotCalibration/Logic/vtkSlicerPivotCalibrationLogic.cxx
```diff
@@ -427,3 +404,80 @@ void vtkSlicerPivotCalibrationLogic::GetToolTipToToolMatrix( vtkMatrix4x4* matri
...
+  if ( fabs( angle ) < vtkMath::RadiansFromDegrees( PARALLEL_ANGLE_THRESHOLD_DEGREES ) ) // If shaft axis and orthogonal axis are not parallel
+  {
+    return vnl_vector< double >( 3, 3, BACKUP_AXIS );
+  }
+  return orthogonalAxis_Shaft;
+}
+
+
+//---------------------------------------------------------------------------
+void vtkSlicerPivotCalibrationLogic::VerifyShaftDirection()
```
lassoan: This method doesn't just verify the direction and returns a status but actually changes the direction. So, it should be rename to something like UpdateShaftDirection.

PivotCalibration/Logic/vtkSlicerPivotCalibrationLogic.h
```diff
@@ -91,6 +95,13 @@ class VTK_SLICER_PIVOTCALIBRATION_MODULE_LOGIC_EXPORT vtkSlicerPivotCalibrationL
   // Computes the maximum orientation difference in degrees between the first tool transformation
   // and all the others. Used for determining if there was enough variation in the input data.
   double GetMaximumToolOrientationDifferenceDeg();
+
+  // Check if the shaft direction is ok, and flip it if necessary
```
lassoan: "flip it if necessary" is not very clear. Could you describe what will be the direction after calling this method (or refer to the place where it is specified)?

PivotCalibration/Logic/vtkSlicerPivotCalibrationLogic.h
```diff
@@ -69,10 +69,14 @@ class VTK_SLICER_PIVOTCALIBRATION_MODULE_LOGIC_EXPORT vtkSlicerPivotCalibrationL
   // Returns with false on failure
   bool ComputeSpinCalibration( bool snapRotation = false ); //Note: The neede orientation protocol assumes that the shaft of the tool lies along the positive x-axis
 
+  // Flip the direction of the shaft axis
+  void FlipShaftDirection();
```
lassoan: Could you add a few lines (probably in the general description of the class) that explains the link between the spin and pivot calibration and shaft direction and also if there is any difference in what order the two calibrations are performed?

lassoan: It would be also helpful to add a few sentences to the module help text about the two types of calibration and in what order they can or should be performed.

PivotCalibration/Resources/UI/qSlicerPivotCalibrationModule.ui
```diff
@@ -152,18 +150,43 @@
...
+        <layout class="QVBoxLayout" name="verticalLayout_3">
+         <item>
+          <widget class="QCheckBox" name="snapCheckBox">
+           <property name="text">
+            <string>Snap rotation to right-angle</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QPushButton" name="flipButton">
```
lassoan: Please add some explanation to the button tooltip about when this button should be used.

lassoan:
Looks good, thank you Matthew.

PR #84, opened 2016-06-07
Add convex hull and always draw volume

aniqah:
1. Adds a convex hull to the output volume that can be toggled with a checkbox.
2. Creates additional points surrounding placed fiducials in the case that the fiducials are on the same plane, and uses these new points to draw a volume. This also means that a volume can be drawn with only <4 and >0 points.


MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -55,6 +61,7 @@ vtkStandardNewMacro(vtkSlicerMarkupsToModelLogic);
 vtkSlicerMarkupsToModelLogic::vtkSlicerMarkupsToModelLogic()
 {
   ImportingScene=0;
```
lassoan: use this-> prefix for all member variables to distinguish them from local variables:
instead of 
ImportingScene = 0;
this->ImportingScene = 0;

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -266,24 +273,21 @@ void vtkSlicerMarkupsToModelLogic::UpdateSelectionNode( vtkMRMLMarkupsToModelNod
 //------------------------------------------------------------------------------
 void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsToModelNode* markupsToModelModuleNode)
 {
-  vtkMRMLMarkupsFiducialNode* markups = markupsToModelModuleNode->GetMarkupsNode(); 
+  vtkMRMLMarkupsFiducialNode* markups = markupsToModelModuleNode->GetMarkupsNode();
   if(markups == NULL)
   {
-    vtkWarningMacro("No markups yet");
```
lassoan: It's OK to remove the warning if this is a normal case, but add a code comment there explaining why it's OK to just return.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -266,24 +273,21 @@ void vtkSlicerMarkupsToModelLogic::UpdateSelectionNode( vtkMRMLMarkupsToModelNod
...
     return;
   }
   int numberOfMarkups = markups->GetNumberOfFiducials();
 
-  if(numberOfMarkups< MINIMUM_MARKUPS_CLOSED_SURFACE_NUMBER)
+  if(numberOfMarkups < 1)
   {
-    //vtkWarningMacro("Not enough fiducials for closed surface");
```
lassoan: replace by comment if you remove the warning (you might also consider debug-level log)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -266,24 +273,21 @@ void vtkSlicerMarkupsToModelLogic::UpdateSelectionNode( vtkMRMLMarkupsToModelNod
...
     if(markupsToModelModuleNode->GetModelNode()!=NULL)
     {
       vtkSmartPointer< vtkSphereSource > sphereSource = vtkSmartPointer< vtkSphereSource >::New();
       markupsToModelModuleNode->GetModelNode()->GetPolyData()->Reset();
       sphereSource->SetRadius(0.00001);
```
lassoan: I'm not sure why we need a data source here (but I guess in order to avoid having VTK warnings somewhere in the pipeline). If we need data source I would prefer something simpler than a sphere. Maybe just a single point. You don't even need to create a source for that, just create a polydata and put a point in it.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -292,45 +296,98 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsT
...
+    markups->GetNthFiducialPosition(i, markupPoint);
+    modelPoints->SetPoint(i, markupPoint);
+    vtkMath::Add(meanPoint, markupPoint, meanPoint);
+    coords[i][0] = markupPoint[0];
+    coords[i][1] = markupPoint[1];
+    coords[i][2] = markupPoint[2];
     modelCellArray->InsertCellPoint(i);
   }
+  vtkMath::MultiplyScalar(meanPoint, 1.0 / numberOfMarkups);
+
```
lassoan: I don't understand easily what you are computing and why. Could you please comment it better? Maybe also move the computation into a separate method.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -292,45 +296,98 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsT
...
+  for (int i = 0; i < numberOfMarkups; i++)
+  {
+    double x1 = coords[i][0];
+    double y1 = coords[i][1];
+    double z1 = coords[i][2];
+    double distance = std::abs(A*x1 + B*y1 + C*z1 + D) / std::sqrt(A*A + B*B + C*C);
+    if (distance >= MINIMUM_THICKNESS)
+    {
+      OpenSurface = false;
```
lassoan: "Open surface" is not the same as "surface points are all on the same plane" (that would be "planar surface" or something like that). If you mean the latter then rename the member variable.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -292,45 +296,98 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsT
...
-    vtkSmartPointer< vtkCleanPolyData > cleanPointPolyData = vtkSmartPointer< vtkCleanPolyData >::New();
-#if (VTK_MAJOR_VERSION <= 5)
-    cleanPointPolyData->SetInput(pointPolyData);
-#else
-    cleanPointPolyData->SetInputData(pointPolyData);
-#endif
-;
-    cleanPointPolyData->SetTolerance(CLEAN_POLYDATA_TOLERANCE);
-    delaunay->SetInputConnection(cleanPointPolyData->GetOutputPort()); //TODO SET VTK5
+    vtkSmartPointer<vtkCubeSource> cube = vtkSmartPointer<vtkCubeSource>::New();
```
lassoan: Cube source was just a quick workaround. Instead please add just two points at a small distance, along a line orthogonal to the fitted plane.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -292,45 +296,98 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsT
...
+    {
+      vtkSmartPointer< vtkCleanPolyData > cleanPointPolyData = vtkSmartPointer< vtkCleanPolyData >::New();
 #if (VTK_MAJOR_VERSION <= 5)
-    delaunay->SetInput( pointPolyData );
+      cleanPointPolyData->SetInput(pointPolyData);
 #else
-    delaunay->SetInputData(pointPolyData);
+      cleanPointPolyData->SetInputData(pointPolyData);
 #endif
+      cleanPointPolyData->SetTolerance(CLEAN_POLYDATA_TOLERANCE);
```
lassoan: all tolerances should be made relative to the point cloud size

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -348,13 +405,24 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsT
     modelNode = markupsToModelModuleNode->GetModelNode();
   }
 
-  if(markupsToModelModuleNode->GetButterflySubdivision())
+  if(markupsToModelModuleNode->GetButterflySubdivision() and not OpenSurface)
   {
     vtkSmartPointer< vtkButterflySubdivisionFilter > subdivisionFilter = vtkSmartPointer< vtkButterflySubdivisionFilter >::New();
```
lassoan: Creating and destroying these numerous filters on each update is a lot of unnecessary work. Please have a look at other logic classes how a structure is created for each node that stores all the filters so that they can be reused for each updated instead of rebuilding the entire pipeline from scratch on each update.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.h
```diff
@@ -206,8 +208,8 @@ vtkMRMLMarkupsToModelNode
   bool CleanMarkups;
   bool ButterflySubdivision;
   double DelaunayAlpha;
+  bool ConvexHull;
```
lassoan: Do you save this in the scene and reload the value from the scene? (there are read/write XML methods in the MRML node cxx)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: The warning was already in UpdateSelectionNode (line 212). Having it here as well meant that the warning would show up twice if there were no markups. It makes more sense to have the one warning in UpdateSelectionNode than a warning in each update type (UpdateOutputCloseSurfaceModel, UpdateOutputCurveModel, etc).

lassoan: Just add this explanation to the code as a comment (maybe compress it to a shorter sentence)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: The warning was already commented-out, I just removed the line. It is no longer necessary at all, because now even one fiducial is enough to draw something. Previously, it would not even attempt to draw a model if there were less than 4 points, because then they must be on the same plane.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: I don't know why this is here either - it was here before. I agree that it seems silly to create a sphere source that is meant to be pretty much invisible. I will change this.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: Okay.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: Good point, I will!

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: Okay.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: Why? Isn't the point placement error independent of the tumor size?

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: Okay.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.h
aniqah: No I didn't, thank you!

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
lassoan: It would be nice to keep this module generic, so that it works reasonably for objects that are 1m, 1mm, 1 um. Size of the point cloud just gives an approximate scale.
I think point placement error may be somewhat fixed in screen coordinates, so it may be an option to get the scale from the viewer's zoom factor, but I think it may be simpler to just use the point cloud extent to get an approximate scale.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: I don't understand. You're saying to add a comment explaining a line that's no longer there?

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: Okay.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
lassoan: Just describe when markup==NULL happens, why it's normal, and why it's OK to just return when it happens. This makes it clear for future developers why this decision has been made and it also forces you to think about it a bit if it's really the correct behavior to just return.

For example: if the user has a model generated from markups and then changes the MarkupsToModel node so that the markups node is NULL then we have to make sure that the model node is cleared (doesn't have any points or cells).

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
aniqah: This tolerance is already relative to the point cloud size.

lassoan: OK then. Thanks for the clarification.

lassoan:
Thank you for the updates!

PR #86, opened 2016-06-23
Add compute normals filter

aniqah:
Adds a new filter option that computes normals for the mesh. The different filter options are placed in a tabbed layout and selected with radio buttons instead of conflicting checkboxes. Basically this is just a new option, so that instead of the butterfly subdivision filter they can select this one, or neither.

### DIFF
--- a/MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
+++ b/MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
@@ -409,35 +409,32 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsT
     modelNode = markupsToModelModuleNode->GetModelNode();
   }
 
-  if(markupsToModelModuleNode->GetButterflySubdivision())
+  if(markupsToModelModuleNode->GetSmoothingType() == 1 || this->PlanarSurface)
   {
-    if (this->PlanarSurface)
+    surfaceFilter->Update();
+    vtkSmartPointer<vtkPolyDataNormals> normals = vtkSmartPointer<vtkPolyDataNormals>::New();
+    normals->SetInputConnection(surfaceFilter->GetOutputPort());
+    normals->SetFeatureAngle(100);
+    normals->Update();
+    modelNode->SetPolyDataConnection(normals->GetOutputPort());
+  }
+  else if(markupsToModelModuleNode->GetSmoothingType() == 2)
+  {
+    vtkSmartPointer< vtkButterflySubdivisionFilter > subdivisionFilter = vtkSmartPointer< vtkButterflySubdivisionFilter >::New();
+    subdivisionFilter->SetInputConnection(surfaceFilter->GetOutputPort());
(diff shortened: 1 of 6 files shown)


lassoan:
vtkPolyDataNormals only computes surface normal, doesn't actually smooth the polygon (the polydata may look simpler because is normals are used for rendering to produce a smoother look of the surface, but that's just how it looks - geometry is not changed). We can add that at the end of the smoothing pipeline for all filters to improve how the surface looks - I'm not sure it is necessary to make it optional.

Why do you think it's useful to make it optional? Is there any use case when it is better not to compute the normals?

tavaughan:
Ok, I'm looking at these changes and how they would integrate with other changes. Could you please clarify what is the purpose of the tabbed widget? Why would the user need to access the closed surface model options when they are using curve mode (or vice versa)? I'm just wondering the reasoning for this decision.

aniqah:
@lassoan: Yes, that is exactly the point! :) The idea is that this will allow the user to make the model *appear* smooth due to shading tricks, but without changing its shape. In contrast, the butterfly subdivision filter will change the model's geometry, which is not always desired.

@tavaughan: I couldn't fit all the options on the screen otherwise. How do you think I should do it instead?

lassoan:
Yes, computing normals and interpolate them during rendering improves how the surface looks. Slicer expects these surface normals to be computed and thus I think it is not necessary to make the computation optional. The user can choose how these values are used for display in the model's display properties (you can choose flat or interpolated normals).

aniqah:
To confirm, you're saying that this step should not be an option, it should always be done?

lassoan:
Yes, exactly. Always compute normal at the end of the pipeline.

ungi:
I would reconsider making it optional. E.g. if I want to visualize a tetrahedron from just four points, maybe I don't want to make it look like a sphere. Also in general, you probably cannot set the feature angle to a value that satisfies all use cases. Some people may want to see 90 degrees as sharp edge, some may prefer to hide it and compute normals there too. I would make compute normals optional, and feature angle an optional parameter.

lassoan:
How to use the surface normals for rendering is a display option. You can choose flat or shaded display in the Models module's Display section (or in the model's display node properties).

tavaughan:
Regarding the tabbed GUI, the cleanest way would probably be to hide non-relevant functionality from the user (e.g. hide curve options when using closed surface mode).

I'm working on a more organized GUI. Unfortunately it's not ready for deployment yet (and won't be until I get back from Germany next week), and I don't think Git will provide a nice way to integrate our combined changes. I suspect we just have to work separately for now and worry about integration later. Maybe someone has a better idea.

ungi:
Would you make the feature angle parameter an option? If not, what would be the default value?

aniqah:
@tavaughan: More organised GUI would be great! I'm not very good at the GUI design yet. Mostly I was trying to avoid the issue where, by showing and hiding options, there is a lot of blank space yet it's difficult to fit everything on the screen.

lassoan:
Feature angle is usually not exposed to the user. In general, the default value is used.

lassoan:
I think this pull request https://github.com/SlicerIGT/SlicerIGT/pull/90 supersedes this one.
@aniqah can I close this pull request?

PR #88, opened 2016-07-03
Better Nrrd support

Utsavvy:
Without this change, plus server stops and complains about not having the correct file writer, since the timestamp becomes part of the extension


lassoan:
Thank you!

PR #90, opened 2016-07-13
Markupstomodel guiupdate

tavaughan:
Not ready for actual merging yet, but please review:
1. How the GUI and MRML node are synchronized
2. Why the combobox selections aren't loaded from saved scenes
Thanks,
Thomas


MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -283,13 +281,14 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsT
   }
   int numberOfMarkups = markups->GetNumberOfFiducials();
 
-  if(numberOfMarkups < 1)
+  if( numberOfMarkups < 1 )
   {
-    if(markupsToModelModuleNode->GetModelNode()!=NULL)
+    if ( markupsToModelModuleNode->GetModelNode() != NULL && markupsToModelModuleNode->GetModelNode()->GetPolyData() != NULL )
     {
+      // TODO: Review this code, see if it is really needed...
```
lassoan: Explain here that we need to set something in the model node's polydata that hides the model but does not generate any warnings.
A small sphere is not ideal, because even if it's hardly visible, it may still throw off "fit-to-screen" function.
Try to set NULL in the polydata and see what happens (check if any warnings are logged).

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -545,7 +544,7 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
   vtkSmartPointer< vtkMRMLModelNode > modelNode;
     modelNode = markupsToModelModuleNode->GetModelNode();
 
-  int totalNumberOfPoints = markupsToModelModuleNode->GetNumberOfIntermediatePoints()*markupsPointsPolyData->GetNumberOfPoints();
+  int totalNumberOfPoints = markupsToModelModuleNode->GetTubeResolutionLength()*markupsPointsPolyData->GetNumberOfPoints();
```
lassoan: This "resolution length" name is quite confusing. Could you find a better one? TubeSamplingDistance, TubeResolution...

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -48,19 +36,19 @@ vtkMRMLMarkupsToModelNode::vtkMRMLMarkupsToModelNode()
...
+  this->AddNodeReferenceRole( OUTPUT_MODEL_ROLE );
+
+  this->AutoUpdateOutput = true;
+  this->CleanMarkups = true;
+  this->ConvexHull = true;
+  this->ButterflySubdivision = true;
+  this->DelaunayAlpha = 0.0;
+  this->TubeRadius = 1.0;
+  this->TubeResolutionLength = 20;
+  this->TubeResolutionAround = 20;
```
lassoan: TubeResolutionAround is not a very clear name. Around what? Can you find a better name?

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -198,9 +201,9 @@ void vtkMRMLMarkupsToModelNode::ReadXMLAttributes( const char** atts )
       nameString << attValue;
       nameString >> this->KochanekTension;
     }
-
   }
-  //vtkDebugMacro("XML atts number of tools read "<<GetNumberOfTools());
+
+  this->Modified();
```
lassoan: Instead of hardcoding calling of Modified(), use Set... functions, which take care of calling Modified() if something is actually modified. Call StartModify/EndModify to prevent multiple Modified event invokations (see for example how it is done in vtkMRMLSegmentEditorNode).

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -214,48 +217,26 @@ void vtkMRMLMarkupsToModelNode::PrintSelf( ostream& os, vtkIndent indent )
...
+{
+  vtkMRMLModelNode* modelNode = vtkMRMLModelNode::SafeDownCast( this->GetNodeReference( OUTPUT_MODEL_ROLE ) );
+  return modelNode;
+}
 
 void vtkMRMLMarkupsToModelNode::SetAndObserveMarkupsNodeID( const char* markupsId )
 {
   // SetAndObserveNodeReferenceID does not handle nicely setting of the same
   // node (it should simply ignore the request, but it adds another observer instead)
   // so check for node equality here.
```
lassoan: Could you check if it's still the case?

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -267,165 +248,318 @@ void vtkMRMLMarkupsToModelNode::SetAndObserveMarkupsNodeID( const char* markupsI
...
   events->InsertNextValue( vtkCommand::ModifiedEvent );
   events->InsertNextValue( vtkMRMLMarkupsNode::PointModifiedEvent ); // PointEndInteractionEvent 
-  this->SetAndObserveNodeReferenceID( MARKUPS_ROLE, markupsId, events.GetPointer() );
+  this->SetAndObserveNodeReferenceID( INPUT_MARKUPS_ROLE, markupsId, events.GetPointer() );
   this->InvokeCustomModifiedEvent(InputDataModifiedEvent);
 }
 
+void vtkMRMLMarkupsToModelNode::SetAndObserveModelNodeID( const char* modelId )
+{
+  // SetAndObserveNodeReferenceID does not handle nicely setting of the same
```
lassoan: Could you check if it's still true?

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -267,165 +248,318 @@ void vtkMRMLMarkupsToModelNode::SetAndObserveMarkupsNodeID( const char* markupsI
...
+    return std::string( this->GetID() ).append( "Model" );
+  }
+  else
+  {
+    return std::string( modelNode->GetName() );
+  }
 }
 
-void vtkMRMLMarkupsToModelNode::RemoveLastMarkup()
+std::string vtkMRMLMarkupsToModelNode::GetModelDisplayNodeName()
```
lassoan: This is such a special method that it should not be part of the MRML node. It belongs more to the GUI.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -267,165 +248,318 @@ void vtkMRMLMarkupsToModelNode::SetAndObserveMarkupsNodeID( const char* markupsI
...
   {
-    markupsNode->RemoveMarkup(markupsNode->GetNumberOfFiducials()-1);
+    return std::string( this->GetID() ).append( "ModelDisplay" );
   }
+
+  return std::string( displayNode->GetName() );
 }
 
-std::string vtkMRMLMarkupsToModelNode::GetModelNodeName()
+std::string vtkMRMLMarkupsToModelNode::GetMarkupsDisplayNodeName()
```
lassoan: This is such a special method that it should not be part of the MRML node. It belongs more to the GUI.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -267,165 +248,318 @@ void vtkMRMLMarkupsToModelNode::SetAndObserveMarkupsNodeID( const char* markupsI
...
-    return std::string(this->GetID()).append("Display");
+    return;
   }
-  else
+  
+  vtkMRMLModelDisplayNode* displayNode = vtkMRMLModelDisplayNode::SafeDownCast( modelNode->GetDisplayNode() );
+  if ( displayNode == NULL )
   {
-    return std::string(this->ModelNode->GetDisplayNode()->GetName());
+    this->createAndObserveModelDisplayNode();
```
lassoan: C has to be capital

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -267,165 +248,318 @@ void vtkMRMLMarkupsToModelNode::SetAndObserveMarkupsNodeID( const char* markupsI
...
   {
-    return std::string(this->ModelNode->GetName());
+    return std::string( this->GetID() ).append( "MarkupsDisplay" );
   }
+
+  return std::string( displayNode->GetName() );
 }
 
-std::string vtkMRMLMarkupsToModelNode::GetDisplayNodeName()
+void vtkMRMLMarkupsToModelNode::SetOutputOpacity( double outputOpacity )
```
lassoan: This kind of utility functions cannot be part of the MRML model node class. The model node should not care about display at all. Such utility functions can be put into the module logic or GUI class.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -267,165 +248,318 @@ void vtkMRMLMarkupsToModelNode::SetAndObserveMarkupsNodeID( const char* markupsI
...
 {
-  if(this->ModelNode == NULL)
+  vtkMRMLModelNode* modelNode = vtkMRMLModelNode::SafeDownCast( this->GetNodeReference( OUTPUT_MODEL_ROLE ) );
```
lassoan: There is lots of copy-paste code for getting/creating the display node. It should get a dedicated method (but not in this class).

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -267,165 +248,318 @@ void vtkMRMLMarkupsToModelNode::SetAndObserveMarkupsNodeID( const char* markupsI
...
+    displayNode->SliceIntersectionVisibilityOn();
+  }
+  else
+  {
+    displayNode->SliceIntersectionVisibilityOff();
   }
 }
 
-void vtkMRMLMarkupsToModelNode::SetOutputIntersectionVisibility(bool outputIntersectionVisibility)
+void vtkMRMLMarkupsToModelNode::SetOutputColor( double redComponent, double greenComponent, double blueComponent )
```
lassoan: All these Set... functions that manipulate the display options should be moved out of this class.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.h
```diff
@@ -31,7 +31,8 @@
 // MarkupsToModel includes
 #include "vtkSlicerMarkupsToModelModuleMRMLExport.h"
 
-static const char* MARKUPS_ROLE = "markupsToModelRoleNode";
+static const char* INPUT_MARKUPS_ROLE = "markupsToModelInputMarkups";
```
lassoan: This role is only used in this class, so no need to prefix it with "markupsToModel". You can just use "InputMarkups" as name.

MarkupsToModel/qSlicerMarkupsToModelModuleWidget.cxx
```diff
@@ -275,391 +245,246 @@ void qSlicerMarkupsToModelModuleWidget::exit()
...
   Q_D( qSlicerMarkupsToModelModuleWidget );
-  qCritical( "PERRAS" );
   this->enter();
   this->updateFromMRMLNode();
-  //this->updateWidget();
 }
 
 //-----------------------------------------------------------------------------
-void qSlicerMarkupsToModelModuleWidget::updateWidget()
+void qSlicerMarkupsToModelModuleWidget::updateToMRMLNode()
```
lassoan: It would be somewhat nicer to have separate callback functions for each widget.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: Actually, removing the code does not seem to have any negative effects. Is there something specific I should be testing for here?

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
tavaughan: Sorry, check if what is still the case?

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.h
tavaughan: It's declared globally in a header file. Could that be an issue?

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
lassoan: Please check if VTK warnings are logged before you add any markups or after you remove all markups and you change display properties of the output model. Since Slicer now uses VTK7, there could have been some changes that made the warnings disappear, so maybe creating an 'almost-empty polydata' is not needed anymore.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
lassoan: I meant that we should check if this is still true:
// SetAndObserveNodeReferenceID does not handle nicely setting of the same  
// node (it should simply ignore the request, but it adds another observer instead)  
// so check for node equality here. 

But I checked the code and actually it is useful to check if the node is changed or not also because it's better not to invoke InputDataModifiedEvent if there is no change. So, just remove this 3-line comment.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.h
lassoan: As it's declared static, the symbols won't be visible in any other files, but it would be nicer to put them into the cxx file.

tavaughan:
I noticed in the console that there is a warning:
"vtkObserverManager ([insert a lot of numbers here]): Observation 8705 between vtkMRMLScene and vtkSlicerMarkupsToModelLogic already exists."
I don't know enough about Slicer's internals to know how to troubleshoot the cause of this, or what any of this means for that matter.

MarkupsToModel/qSlicerMarkupsToModelModuleWidget.cxx
tavaughan: I know that would be slightly better for performance. I don't know that it will be a noticeable improvement though (I don't notice any delay in the current implementation), and it makes the code harder to work with (there are a lot of widgets, which == a lot of new function calls, which == many more lines of code to keep track of). I would prefer to leave it as-is, because right now it's easy to see how everything is updated, all in one place.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
tavaughan: The logic class seems to use it a few times, so I made it a static logic function

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
tavaughan: TubeNumberOfSides

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: TubeSamplingFrequency

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: No warnings related to empty poly data. I did change the logic a bit though to allow empty poly data

tavaughan: Correction - with slice intersections visible, it does give a "No input data" error, but no other errors or crashes.

lassoan: I've submitted a pull request for fixing this "No input data" warning in Slicer core (https://github.com/Slicer/Slicer/pull/563)

PR #93, opened 2016-08-05
Markups to model gui update

tavaughan:


lassoan:
Could you please send a pull request with all the commits squashed into one single commit?

PR #97, opened 2016-08-30
ENH: Added Polynomial curve fitting to MarkupsToModel

tavaughan:


MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -58,7 +55,10 @@
 // STD includes
 #include <cassert>
 #include <cmath>
-#include <vector> // TODO: Evaluate whether needed
+#include <vector>
+#include <set>
+
+#define LINE_MIN_NUMBER_POINTS 2
```
lassoan: use static const int, it's typesafe

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -458,14 +458,25 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputLinearModel(vtkMRMLMarkupsToModel
...
+    vtkErrorMacro("Markups contains a vtkPoints object that is null. No model generated.");
+    return;
+  }
 
-  points = markupsPointsPolyData->GetPoints();
+  int numPoints = points->GetNumberOfPoints();
+  // redundant error checking, to be safe
+  if (numPoints < LINE_MIN_NUMBER_POINTS)
+  {
+    vtkErrorMacro("Not enough points to create an output linear model. Need at least 2, " << numPoints << " are provided. No output created.");
```
lassoan: use LINE_MIN_NUMBER_POINTS in the comment, too

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+
+  int numPoints = points->GetNumberOfPoints();
+  // redundant error checking, to be safe
+  if (numPoints < LINE_MIN_NUMBER_POINTS)
+  {
+    vtkErrorMacro("Not enough points to compute polynomial parameters. Need at least " << LINE_MIN_NUMBER_POINTS << " points but " << numPoints << " are provided.");
+    return;
+  }
+
+  if (markupsPointsParameters->GetNumberOfTuples())
```
lassoan: add >0 to make it more legible

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  }
+
+  for (int v = 0; v < numPoints; v++)
+  {
+    markupsPointsParameters->InsertNextTuple1( v / double(numPoints-1) );
+    // division to clamp all values to range 0.0 - 1.0
+  }
+}
+
+//------------------------------------------------------------------------------
```
lassoan: This class is getting really big. Please move polynomial fitting code to another class. Maybe create a http://www.vtk.org/doc/nightly/html/classvtkParametricFunction.html and move the algorithms into that. You may even contribute it to VTK or as a VTK external module in the future.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  {
+    vtkErrorMacro("Not enough points to compute polynomial parameters. Need at least " << LINE_MIN_NUMBER_POINTS << " points but " << numPoints << " are provided.");
+    return;
+  }
+
+  // vtk boost algorithms cannot be used because they are not built with 3D Slicer
+  // so this is a custom implementation of:
+  // 1. constructing an undirected graph as a 2D array
+  // 2. Finding the two vertices that are the farthest apart
+  // 3. running prim's algorithm on the graph
```
lassoan: describe what Prim's algorithm does (constructs min spanning tree)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+
+  int numPoints = points->GetNumberOfPoints();
+  // redundant error checking, to be safe
+  if (numPoints < LINE_MIN_NUMBER_POINTS)
+  {
+    vtkErrorMacro("Not enough points to compute polynomial parameters. Need at least " << LINE_MIN_NUMBER_POINTS << " points but " << numPoints << " are provided.");
+    return;
+  }
+
+  // vtk boost algorithms cannot be used because they are not built with 3D Slicer
```
lassoan: when you need a long comment like this, it indicates that the method should be split into 5 distinct methods - of course that many additional methods in this already busy class would not be nice => one more reason to move all polyfit related methods into a separate class

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  double maximumDistance = 0;
+  // iterate through all points
+  for (int v = 0; v < numPoints; v++)
+  {
+    for (int u = 0; u < numPoints; u++)
+    {
+      double pointU[3], pointV[3];
+      points->GetPoint(u, pointU);
+      points->GetPoint(v, pointV);
+      double distX = (pointU[0] - pointV[0]); double distXsq = distX * distX;
```
lassoan: don't put multiple statements in one line, it's just not nice style (makes the code difficult to read)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  // 1. constructing an undirected graph as a 2D array
+  // 2. Finding the two vertices that are the farthest apart
+  // 3. running prim's algorithm on the graph
+  // 4. extract the "trunk" path from the last vertex to the first
+  // 5. based on the distance along that path, assign each vertex a polynomial parameter value
+
+  // in the following code, two tasks are done:
+  // 1. construct an undirected graph
+  std::vector<double> distances(numPoints*numPoints);
+  distances.assign(numPoints*numPoints, 0.0);
```
lassoan: assign value in the constructor instead
distances(numPoints*numPoints, 0.0)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  {
+    for (int u = 0; u < numPoints; u++)
+    {
+      double pointU[3], pointV[3];
+      points->GetPoint(u, pointU);
+      points->GetPoint(v, pointV);
+      double distX = (pointU[0] - pointV[0]); double distXsq = distX * distX;
+      double distY = (pointU[1] - pointV[1]); double distYsq = distY * distY;
+      double distZ = (pointU[2] - pointV[2]); double distZsq = distZ * distZ;
+      double dist3D = sqrt(distXsq+distYsq+distZsq);
```
lassoan: use vtkMath::Distance2BetweenPoints (it returns squared distance)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  while (currentPathIndex != -1)
+  {
+    pathIndices.push_back(currentPathIndex);
+    currentPathIndex = parent[currentPathIndex]; // go up the tree one layer
+  }
+    
+  // find the sum of distances along the trunk path of the tree
+  double sumOfDistances = 0.0;
+  for (int i = 0; i < pathIndices.size() - 1; i++)
+    sumOfDistances += graph[i][i+1];
```
lassoan: always add { } after for

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+
+  // check this to prevent a division by zero (in case all points are duplicates)
+  if (sumOfDistances == 0)
+  {
+    vtkErrorMacro("Minimum spanning tree path has distance zero. No parameters will be assigned. Check inputs!");
```
lassoan: "Check inputs!" - message to whom? what inputs? to what? refer to method name and specific arguments that may be wrong

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  double currentDistance = 0.0;
+  for (int i = 0; i < pathIndices.size() - 1; i++)
+  {
+    pathParameters.push_back(currentDistance/sumOfDistances);
+    currentDistance += graph[i][i+1];
+  }
+  pathParameters.push_back(currentDistance/sumOfDistances); // this should be 1.0
+
+  // finally assign polynomial parameters to each point, and store in the output array
+  if (markupsPointsParameters->GetNumberOfTuples())
```
lassoan: add >0

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: I agree that the class is big, and I think there are a lot of reasons for this. I think refactoring the logic should be done as its own separate task, that way we can properly plan which parts belong in the class, and which parts should be in other files. We can also set up the proper vtk pipeline in this task to make the module overall more efficient.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+    vtkWarningMacro("Not enough points to compute a polynomial fit. " << "For an order " << markupsToModelModuleNode->GetPolynomialOrder() << " polynomial, at least " << numPolynomialCoefficients << " points with unique parameters are needed. "
+                    << numUniquePointParameters << " points with unique parameters were found. "
+                    << "An order " << (numUniquePointParameters - 1) << " polynomial will be created instead.");
+    numPolynomialCoefficients = numUniquePointParameters;
+  }
+
+  // independent values (parameter along the curve)
+  int numIndependentValues = numPoints * numPolynomialCoefficients;
+  std::vector<double> independentValues(numIndependentValues); // independent values
+  independentValues.assign(numIndependentValues, 0.0);
```
lassoan: assign initial value in constructor

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: ooh, neat! Didn't know about that constructor. :)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  {
+    independentMatrix[p] = &(independentValues[p * numPolynomialCoefficients]);
+  }
+  double** independentMatrixPtr = &(independentMatrix[0]);
+
+  // dependent values
+  const int numDimensions = 3; // this should never be changed from 3
+  int numDependentValues = numPoints * numDimensions;
+  std::vector<double> dependentValues(numDependentValues); // dependent values
+  dependentValues.assign(numDependentValues, 0.0);
```
lassoan: move initialization to constructor everywhere

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+      independentValues[p * numPolynomialCoefficients + c] = value;
+    }
+  }
+  std::vector<double*> independentMatrix(numPoints);
+  independentMatrix.assign(numPoints, NULL);
+  for (int p = 0; p < numPoints; p++)
+  {
+    independentMatrix[p] = &(independentValues[p * numPolynomialCoefficients]);
+  }
+  double** independentMatrixPtr = &(independentMatrix[0]);
```
lassoan: move variable creation to the closest place it is first _used_ (I had to scroll through a lot to find where this variable is actually used and for what)

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+    coefficientMatrix[c] = &(coefficientValues[c * numDimensions]);
+  }
+  double** coefficientMatrixPtr = &(coefficientMatrix[0]); // the solution
+
+  // Input the forumulation into SolveLeastSquares
+  vtkMath::SolveLeastSquares(numPoints, independentMatrixPtr, numPolynomialCoefficients, dependentMatrixPtr, numDimensions, coefficientMatrixPtr);
+
+  // Use the values to generate points along the polynomial curve
+  vtkSmartPointer<vtkPoints> smoothedPoints = vtkSmartPointer<vtkPoints>::New(); // points
+  vtkSmartPointer< vtkCellArray > smoothedLines = vtkSmartPointer<  vtkCellArray >::New(); // lines
```
lassoan: use spaces around < > consistently

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -538,6 +577,373 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputHermiteSplineModel(vtkMRMLMarkups
...
+  }
+
+  // Convert the points to a tube model
+  vtkSmartPointer< vtkPolyData >smoothedSegments = vtkSmartPointer< vtkPolyData >::New();
+  smoothedSegments->Initialize();
+  smoothedSegments->SetPoints(smoothedPoints);
+  smoothedSegments->SetLines(smoothedLines);
+  
+  vtkSmartPointer< vtkTubeFilter> tubeFilter = vtkSmartPointer< vtkTubeFilter>::New();
+#if (VTK_MAJOR_VERSION <= 5)
```
lassoan: Slicer is no more backward-compatible with VTK5, so you can remove VTK_MAJOR_VERSION<=5 branches

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -86,12 +89,14 @@ void vtkMRMLMarkupsToModelNode::WriteXML( ostream& of, int nIndent )
   of << indent << " ButterflySubdivision =\"" << this->ButterflySubdivision << "\"";
   of << indent << " DelaunayAlpha =\"" << this->DelaunayAlpha << "\"";
   of << indent << " InterpolationType=\"" << this->InterpolationType << "\"";
+  of << indent << " InterpolationType=\"" << this->PointParameterType << "\"";
```
lassoan: copy-paste error

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -86,12 +89,14 @@ void vtkMRMLMarkupsToModelNode::WriteXML( ostream& of, int nIndent )
   of << indent << " ButterflySubdivision =\"" << this->ButterflySubdivision << "\"";
   of << indent << " DelaunayAlpha =\"" << this->DelaunayAlpha << "\"";
   of << indent << " InterpolationType=\"" << this->InterpolationType << "\"";
```
lassoan: convert to string and print that

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -159,6 +164,19 @@ void vtkMRMLMarkupsToModelNode::ReadXMLAttributes( const char** atts )
       }
       SetInterpolationType(interpolationType);
     }
+    else if ( ! strcmp( attName, "PointParameterType" ) )
+    {
+      int pointParameterType = 0;
+      std::stringstream nameString;
+      nameString << attValue;
+      nameString >> pointParameterType;
```
lassoan: Don't store magic numbers in files. How would you know what PointParameterType="1" means? You could document it somewhere but then you have to create and maintain that documentation and make it easy to find for people who want to interpret your files, etc.
Instead, convert the enum to a string and store that in file - as it is done for InterpolationType.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
```diff
@@ -329,8 +360,20 @@ const char* vtkMRMLMarkupsToModelNode::GetInterpolationTypeAsString( int id )
   {
   case Linear: return "linear";
   case CardinalSpline: return "cardinalSpline";
-  case HermiteSpline: return "hermiteSpline";
   case KochanekSpline: return "kochanekSpline";
```
lassoan: That's also a reason why you want to store enums as strings in a file!!!
kochanekSpline has just become 2, while it was 3 before. If you stored "3" in a file as interpolation type, user would be surprised to see that when he loads an earlier file, interpolation type is changed to polynomial.

lassoan:
Thank you for working on this, you've done nice work. There are a couple of things to improve (see in code review comments), but nothing that couldn't wait, so I've merged the pull request.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
tavaughan: I'm a little bit confused - in what way am I treating this variable differently from InterpolationType? The code looks pretty much the same to me...

tavaughan: I do understand why we want to do this, and I will store enum strings instead of ints in the settings file.  I assume I need to make changes to InterpolationType too (?)

tavaughan:
Thanks Andras for the review - I will make the changes you requested.

MarkupsToModel/MRML/vtkMRMLMarkupsToModelNode.cxx
lassoan: I'm sorry, you are right. Github just showed a diff and I didn't see the string<->enum conversion.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: Is there an established standard with regards to spaces around < >, [ ], and ( ) brackets? I've seen both used in different code.

lassoan: There are multiple "standards".

Some compilers require a space between two >s. For example:
std::vector<std::vector<double>> fails on some compilers
std::vector< std::vector<double> > does not cause problems anywhere.
For this reason, probably it is better to get used to putting spaces there. But most importantly, be consistent, at least within any method that you touch (otherwise you make the reader wondering if there is a reason for the different spacing).

PR #105, opened 2016-11-03
ENH: Fixed up TransformFusion, added auto update, added new fusion mode

tavaughan:
There are still some changes needed:
- Should not be able to add the same transform more than once for quaternion average
- Timed update is controlled in the widget

I would like to get this merged first, then deal with those issues separately. Already this change set is... complex.

### DIFF
--- a/TransformFusion/Logic/vtkMRMLTransformFusionNode.cxx
+++ b/TransformFusion/Logic/vtkMRMLTransformFusionNode.cxx
@@ -35,10 +35,10 @@
 //----------------------------------------------------------------------------
 // constant strings for MRML reference roles
 // these should not be used outside of this class
-std::string ROLE_INPUT_TRANSFORM     = std::string( "Input Transform" );
-std::string ROLE_RESTING_TRANSFORM   = std::string( "Resting Transform" );
-std::string ROLE_REFERENCE_TRANSFORM = std::string( "Reference Transform" );
-std::string ROLE_OUTPUT_TRANSFORM    = std::string( "Output Transform" );
+const char* ROLE_INPUT_TRANSFORM     = "Input Transform";
+const char* ROLE_RESTING_TRANSFORM   = "Resting Transform";
+const char* ROLE_REFERENCE_TRANSFORM = "Reference Transform";
+const char* ROLE_OUTPUT_TRANSFORM    = "Output Transform";
 
 //----------------------------------------------------------------------------
 vtkMRMLNodeNewMacro( vtkMRMLTransformFusionNode );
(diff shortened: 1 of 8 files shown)


tavaughan:
@lassoan Did you make comments on this commit? I see from the list of pull request that this has 20 comments... but I cannot find any comments.

lassoan:
@tavaughan you should be able to see my comments now

TransformFusion/Logic/vtkSlicerTransformFusionLogic.cxx
```diff
@@ -38,133 +38,381 @@
...
+    return;
+  }
+
+  if ( event==vtkMRMLTransformFusionNode::InputDataModifiedEvent )
+  {
+    if ( paramNode->GetUpdateMode() == vtkMRMLTransformFusionNode::UPDATE_MODE_AUTO )
     {
-      this->QuaternionAverageFusion();
-      break;
+      this->Fuse( paramNode );
```
tavaughan: Would FuseTransforms work? If not I will just say UpdateOutputTransform

TransformFusion/Logic/vtkSlicerTransformFusionLogic.cxx
```diff
@@ -38,133 +38,381 @@
...
     }
 
-    vtkMath::Matrix3x3ToQuaternion(rotationMatrix, singleQuaternion);
+    vtkMath::Matrix3x3ToQuaternion( rotationMatrix, singleQuaternion );
 
-    averageQuaternion[0] = averageQuaternion[0] + singleQuaternion[0];
-    averageQuaternion[1] = averageQuaternion[1] + singleQuaternion[1];
-    averageQuaternion[2] = averageQuaternion[2] + singleQuaternion[2];
-    averageQuaternion[3] = averageQuaternion[3] + singleQuaternion[3];
+    averageQuaternion[ 0 ] = averageQuaternion[ 0 ] + singleQuaternion[ 0 ];
```
tavaughan: I'll double check later that this is following the algorithm, but here:
http://stackoverflow.com/questions/12374087/average-of-multiple-quaternions

I will add a comment with a link to here if appropriate

TransformFusion/Logic/vtkMRMLTransformFusionNode.cxx
```diff
@@ -52,91 +66,309 @@ vtkMRMLTransformFusionNode::~vtkMRMLTransformFusionNode()
...
+  os << indent << " UpdateMode = " << GetUpdateModeAsString( this->UpdateMode ) << "\n";
+  os << indent << " FusionMode = " << GetFusionModeAsString( this->FusionMode ) << "\n";
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLTransformFusionNode::Copy( vtkMRMLNode *anode )
+{
+  Superclass::Copy( anode );
+  vtkMRMLTransformFusionNode *node = vtkMRMLTransformFusionNode::SafeDownCast( anode );
   this->DisableModifiedEventOn();
```
tavaughan: Could you refer to an example where this is done correctly?

TransformFusion/Logic/vtkSlicerTransformFusionLogic.cxx
lassoan: If the method updates the output transform then UpdateOutputTransform is a better name. If you compute fuse transforms and return it from the method then GetFusedTransform would be a good name. FuseTransforms is not specific enough, it's not clear what happens when you call the method.

TransformFusion/Logic/vtkMRMLTransformFusionNode.cxx
lassoan: Search for StartModify anywhere in Slicer or SlicerIGT, but it's really simple:
  int wasModifying = node->StartModify();
  ...make several modifications to node
  node->EndModify( wasModifying );

TransformFusion/Logic/vtkSlicerTransformFusionLogic.cxx
tavaughan: This was the pre-existing implementation in the module, I personally have no applications for quaternion average. I could comment + document that the existing implementation is incorrect, but that it could be updated to a proper solution: http://www.acsu.buffalo.edu/~johnc/ave_quat07.pdf

lassoan: Documenting what we know in a comment would be enough, thank you.

tavaughan:
I believe I've addressed all of your comments

lassoan:
Thank you!

PR #110, opened 2017-01-12
ENH: Added ctkPathLineEdit widget to Guidelet Settings

zacbaum:
Modified the GUI in the settings panel of the guidelet to have a
ctkPathLineEdit widget instead of a QLineEdit so that users can browse
through to choose their save-to directory instead of manually entering
it.


lassoan:
There are conflicts that I don't have time to resolve right now, but if you resolve and resubmit the pull request then I'll merge it quickly.

PR #111, opened 2017-01-12
BUG: When a guidelet is reloaded from the module widget, the guidelet…

adamrankin:
… was not properly cleaned up, as the dock widget remained.

The guidelet is now fully cleaned up.


lassoan:
Thank you!

PR #112, opened 2017-01-12
Markups to model ring mode

tavaughan:
These are enhancements that improve both the predictability and flexibility of curve fitting in MarkupsToModel.
- Interpolating curves will now pass through the control points (as they should have)
- There are additional options that allow this module to better emulate the original behaviour of CurveMaker


lassoan:
Thank you, it looks good, I merge this as is to make it available in the nightly builds. You can further refine alpha shapes pipeline and send another pull request.

PR #113, opened 2017-01-13
ENH: Code format cleaning.

adamrankin:
Guaranteeing existence of ReferenceToRAS in UltraSound.setupScene by moving it to Guidelet class.

Renaming file to match class name (GuideletFrame->Guidelet)

### DIFF
--- a/Guidelet/GuideletLib/Guidelet.py
+++ b/Guidelet/GuideletLib/Guidelet.py
@@ -6,7 +6,6 @@ import time
 from UltraSound import UltraSound
 
 class Guidelet(object):
-
   @staticmethod
   def showToolbars(show):
 
@@ -22,24 +21,10 @@ class Guidelet(object):
     except:
       pass
 
-  def showModulePanel(self, show):
-    modulePanelDockWidget = slicer.util.mainWindow().findChildren('QDockWidget','PanelDockWidget')[0]
-    modulePanelDockWidget.setVisible(show)
-
-    if show:
-      mainWindow=slicer.util.mainWindow()
-      if self.sliceletDockWidgetPosition == qt.Qt.LeftDockWidgetArea:
-        mainWindow.tabifyDockWidget(self.sliceletDockWidget, modulePanelDockWidget)
-        self.sliceletDockWidget.setFeatures(qt.QDockWidget.DockWidgetClosable+qt.QDockWidget.DockWidgetMovable+qt.QDockWidget.DockWidgetFloatable)
-    else:
-      if self.sliceletDockWidgetPosition == qt.Qt.LeftDockWidgetArea:
-        # Prevent accidental closing or undocking of the slicelet's left panel
-        self.sliceletDockWidget.setFeatures(0)
-
   @staticmethod
   def showPythonConsole(show):
     slicer.util.mainWindow().pythonConsole().parent().setVisible(show)
-        
+
   @staticmethod
   def showMenuBar(show):
     for menubar in slicer.util.mainWindow().findChildren('QMenuBar'):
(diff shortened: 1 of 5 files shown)


Guidelet/GuideletLib/Guidelet.py
```diff
@@ -297,18 +306,41 @@ def registerCustomLayouts(self):#common
...
 
+  # -----------------------------------------------------
+  def onSceneLoaded(self):
+    # Derived classes can override this function
+    pass
 
+  # -----------------------------------------------------
   def setupScene(self):
+    '''
+    ReferenceToRas transform is used in almost all IGT applications. Reference is the coordinate system
```
lassoan: Guidelet does not need and does not know about ReferenceToRas. Please move this to UltraSound.setupScene().

Guidelet/GuideletLib/Guidelet.py
```diff
@@ -297,18 +306,41 @@ def registerCustomLayouts(self):#common
...
+    of a tool fixed to the patient. Tools are tracked relative to Reference, to compensate for patient
+    motion. ReferenceToRas makes sure that everything is displayed in an anatomical coordinate system, i.e.
+    R, A, and S (Right, Anterior, and Superior) directions in Slicer are correct relative to any
+    images or tracked tools displayed.
+    '''
+    self.referenceToRas = slicer.util.getNode('ReferenceToRas')
```
lassoan: If self.referenceToRas exists and it is valid then just use that (don't overwrite it with slicer.util.getNode('ReferenceToRas')  then).

adamrankin:
Ok, it's all cleaned up, referenceToRas only lives in the Ultrasound class.

Guidelet/GuideletLib/UltraSound.py
```diff
@@ -4,12 +4,13 @@
 import time
 
 class UltraSound(object):
-
   DEFAULT_IMAGE_SIZE = [800, 600, 1]
 
+  # -----------------------------------------------------
```
lassoan: We don't add separator lines as text to py files, because IDE's can draw separator lines automatically (and having the separators as text would also make code folding uglier - instead of seeing only the methods, there would be always that manual separator line displayed as well). What Python IDE do you use?

adamrankin: PyCharm

Guidelet/GuideletLoadable.py
```diff
@@ -265,13 +281,14 @@ class GuideletTest(ScriptedLoadableModuleTest):
   """
   #TODO add common test utility methods here
 
+  # -----------------------------------------------------
   def setUp(self):
-    """ Do whatever is needed to reset the state - typically a scene clear will be enough.
-    """
+    # Do whatever is needed to reset the state - typically a scene clear will be enough.
```
adamrankin: Oh, is """ special syntax for doxygen?

Guidelet/GuideletLib/UltraSound.py
lassoan: In PyCharm: File / Settings / Editor / General / Appearance / Show method separators.
You can configure it to have a stronger color: File / Settings / Editor / General / Editor / Colors & Fonts / General / Code / Method separator color.

Guidelet/GuideletLoadable.py
lassoan: Python standard method documentation format (https://www.python.org/dev/peps/pep-0257/).
Try for example this in the Slicer console:
help(slicer.util.getNode)

Guidelet/GuideletLib/UltraSound.py
adamrankin: Nice! I will remove the code separators.

Guidelet/GuideletLoadable.py
adamrankin: Very neat. I will revert them to method docs

adamrankin:
Ok, I think it's pruned down to just real changes

PR #119, opened 2017-04-10
Added Recording Always Enabled Parameter.

zacbaum:
If set to True (False by default in Guidelet base) in the user's
guidelet parameters, the recording button will remain clickable even if
the plus connector is frozen.

### DIFF
--- a/Guidelet/GuideletLoadable.py
+++ b/Guidelet/GuideletLoadable.py
@@ -191,7 +191,7 @@ class GuideletLogic(ScriptedLoadableModuleLogic):
                    'RecordingFilenameExtension' : '.mhd',
                    'SavedScenesDirectory' : defaultSavePath,
                    'UltrasoundBrightnessControl' : 'Buttons',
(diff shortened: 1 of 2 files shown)


Guidelet/GuideletLoadable.py
```diff
@@ -190,7 +190,8 @@ def addValuesToDefaultConfiguration(self):
                    'RecordingFilenamePrefix' : 'GuideletRecording-',
                    'RecordingFilenameExtension' : '.mhd',
                    'SavedScenesDirectory' : defaultSavePath,
-                   'UltrasoundBrightnessControl' : 'Buttons'
+                   'UltrasoundBrightnessControl' : 'Buttons',
+                   'RecordingAlwaysEnabled' : 'False',
```
lassoan: "Always" is too general. We may still want to make it disabled in certain cases. Change it to: "RecordingEnabledWhenConnectorNodeDisconnected".

PR #123, opened 2017-05-11
Slicerigt general fixes

tavaughan:


MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.h
```diff
@@ -111,6 +107,29 @@ class VTK_SLICER_MARKUPSTOMODEL_MODULE_LOGIC_EXPORT vtkSlicerMarkupsToModelLogic
   vtkSlicerMarkupsToModelLogic(const vtkSlicerMarkupsToModelLogic&); // Not implemented
   void operator=(const vtkSlicerMarkupsToModelLogic&); // Not implemented
   int ImportingScene;
+
+  enum PointArrangementEnum
```
lassoan: "enum" is usually not included in the type name

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -59,7 +61,11 @@
 #include <vector>
 #include <set>
 
-#define LINE_MIN_NUMBER_POINTS 2
+// local constants
+static const int LINE_MIN_NUMBER_POINTS = 2;
+static const int MINIMUM_MARKUPS_NUMBER = 3;
```
lassoan: We usually follow VTK convention of calling "count" variables as NumberOfX. So, it should be something like NUMBER_OF_MARKUPS_MIN and NUMBER_OF_LINE_POINTS_MIN.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -59,7 +61,11 @@
...
+static const double CLEAN_POLYDATA_TOLERANCE = 0.01;
```
lassoan: unit? if no unit then ratio of what values?

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -276,135 +282,147 @@ void vtkSlicerMarkupsToModelLogic::UpdateSelectionNode( vtkMRMLMarkupsToModelNod
...
 
 //------------------------------------------------------------------------------
-void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsToModelNode* markupsToModelModuleNode, vtkPolyData* output)
+void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel( vtkMRMLMarkupsToModelNode* markupsToModelModuleNode, vtkPolyData* output )
 {
   vtkMRMLMarkupsFiducialNode* markups = markupsToModelModuleNode->GetMarkupsNode();
-  if(markups == NULL)
+  if ( markups == NULL )
   {
     return;
```
lassoan: don't you need to clean the displayed surface model?

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -276,135 +282,147 @@ void vtkSlicerMarkupsToModelLogic::UpdateSelectionNode( vtkMRMLMarkupsToModelNod
...
   }
   int numberOfMarkups = markups->GetNumberOfFiducials();
 
-  if (numberOfMarkups == 0)
+  if ( numberOfMarkups == 0 )
   {
     return;
```
lassoan: don't you need to clean the displayed surface model?

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -276,135 +282,147 @@ void vtkSlicerMarkupsToModelLogic::UpdateSelectionNode( vtkMRMLMarkupsToModelNod
...
+    inputPoints->SetPoint( i, markupPoint );
+  }
 
-  for (int i = 0; i < numberOfMarkups; i++)
+  vtkSmartPointer< vtkPolyData > inputPolyData = vtkSmartPointer< vtkPolyData >::New();
+  inputPolyData->SetLines( inputCellArray );
+  inputPolyData->SetPoints( inputPoints );
+  
+  vtkSmartPointer< vtkCleanPolyData > cleanPointPolyData = vtkSmartPointer< vtkCleanPolyData >::New();
+  #if (VTK_MAJOR_VERSION <= 5)
```
lassoan: No need for VTK<=5 support anymore

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -435,6 +453,225 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel(vtkMRMLMarkupsT
...
+    return;
+  }
+  outputPlaneNormal[ 0 ] = variationMinimumOBBAxis[ 0 ];
+  outputPlaneNormal[ 1 ] = variationMinimumOBBAxis[ 1 ];
+  outputPlaneNormal[ 2 ] = variationMinimumOBBAxis[ 2 ];
+  vtkMath::Normalize( outputPlaneNormal );
+}
+
+//------------------------------------------------------------------------------
+vtkSlicerMarkupsToModelLogic::PointArrangementEnum vtkSlicerMarkupsToModelLogic::ComputePointArrangement( vtkPoints* points, double planeNormal[ 3 ], double lineAxis[ 3 ] )
```
lassoan: ComputePointArrangement, ComputePointRangeAlongAxis, ComputeSurfaceExtrusionAmount are way too complicated.
ComputeBestFitPlaneAxes should just always return all three axes as a vtkMatrix4x4 (x axis: largest variance; z axis: smallest variance). You then use this matrix to transform all points (transform polydata filter) to the bounding box coordinate system and get the extents (output->GetExtents).

If extentX<eps => singular;
else if extentY/extentX<eps => linear;
else if extentZ/sqrt(extentX*extentX+extentY*extent) < someThreshold => planar;
else nonplanar.

Extrusion amount: 1/100 of the norm of the bounding box extent. If extent is 0 then I think a fixed value is OK.

tavaughan:
I just reinstalled Windows so it may be a few days of setup before I can get to this.

Tamas and Valeria wanted this fix so they could proceed with their research. This should probably be merged now.

We should create a separate issue so I remember to address your comments in code.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: That's... great to hear!

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: Extrusion amount was determined through testing. I think I wrote something to that effect in a comment, but perhaps I need to clarify:

The problem is if the extrusion amount is reduced below about 1/10 of the maximum extend, then the delaunay filter can fail and not produce any output. So you end up with weird "gray zones" between planar and non-planar surfaces where you get no output at all.

lassoan:
Thank you Thomas. As you suggested, I've merged this as is, and added https://github.com/SlicerIGT/SlicerIGT/issues/125 to address the comments later.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: What do you mean by "clean"?

lassoan: By "clean the displayed model" I meant remove all points in the model (to not show anything).
When there are no input markups then we should make sure we do not display a surface (previously there could have been markups that generated a surface).

tavaughan: I think that bug was removed a while ago, but I agree that this method should make that more explicit. I'll call vtkPolyData::Reset().

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: The proposed approach involves the inclusion of four additional header files. I'm not sure if this is consistent with minimizing cohesion to other classes. Unless I hear otherwise, I will proceed under the assumption that this is fine:
- vtkMatrix4x4
- vtkTransform
- vtkTransformFilter
- vtkGeometryFilter (to compute extents)

lassoan: These are fine. These are very generic classes with a well-defined interface that are only used temporarily in certain methods, which results in a very weak unidirectional association.

tavaughan: Ok. While we're here :-), I'm having some trouble getting the extent from a polydata. I have code:

  vtkPoints* transformedPoints = transformFilter->GetPolyDataOutput()->GetPoints();
  transformedPoints->ComputeBounds();
  double* extents = transformedPoints->GetBounds(); // { xmin, xmax, ymin, ymax, zmin, zmax }

But the elements are always 0. I've also tried getting the bounds from the vtkPolyData output directly. Am I missing a step somewhere? I can confirm that there are in fact points (as determined by calling GetNumberOfPoints)...

tavaughan: Disregard question, I think I figured it out. :-)

Issue #125, opened 2017-05-15
Simplify MarkupsToModel singular surface detection and compensation


lassoan:
This is a follow-up on this pull request: https://github.com/SlicerIGT/SlicerIGT/pull/123#issuecomment-301540958

ComputePointArrangement, ComputePointRangeAlongAxis, ComputeSurfaceExtrusionAmount are way too complicated.
ComputeBestFitPlaneAxes should just always return all three axes as a vtkMatrix4x4 (x axis: largest variance; z axis: smallest variance). You then use this matrix to transform all points (transform polydata filter) to the bounding box coordinate system and get the extents (output->GetExtents).
If extentX<eps => singular;
else if extentY/extentX<eps => linear;
else if extentZ/sqrt(extentXextentX+extentYextent) < someThreshold => planar;
else nonplanar.
Extrusion amount: 1/100 of the norm of the bounding box extent. If extent is 0 then I think a fixed value is OK.

tavaughan:
Thanks. I'll take a look when I get the chance.

PR #126, opened 2017-05-26
STY: Addressing some outstanding style in MarkupsToModel

tavaughan:


MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -284,16 +286,36 @@ void vtkSlicerMarkupsToModelLogic::UpdateSelectionNode( vtkMRMLMarkupsToModelNod
...
+    vtkErrorMacro( "No output poly data provided to UpdateOutputCloseSurfaceModel! No operation performed!" );
     return;
   }
+
+  output->Reset(); // empty the poly data of all existing data
+
+  bool outputContainsData = output->GetNumberOfPoints();
+  if ( outputContainsData )
+  {
+    vtkWarningMacro( "Output poly data provided to UpdateOutputCloseSurfaceModel already contains data. Existing data will be overwritten." );
```
lassoan: This is normal, no need to log warning

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -284,16 +286,36 @@ void vtkSlicerMarkupsToModelLogic::UpdateSelectionNode( vtkMRMLMarkupsToModelNod
...
+
   vtkMRMLMarkupsFiducialNode* markups = markupsToModelModuleNode->GetMarkupsNode();
-  if ( markups == NULL )
+  if ( output == NULL )
   {
+    vtkErrorMacro( "No output poly data provided to UpdateOutputCloseSurfaceModel! No operation performed!" );
     return;
   }
+
+  output->Reset(); // empty the poly data of all existing data
```
lassoan: Memory allocation/deallocation is expensive in general. If it doesn't make the code more complicated, it's better to do it only when needed (e.g., in the branch where you return if there are input markups).

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -333,23 +351,33 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel( vtkMRMLMarkups
 
   vtkSmartPointer< vtkPolyData > cleanedPolyData = vtkSmartPointer< vtkPolyData >::New();
```
lassoan: No need for creating a smart pointer for this, just use a simple pointer. Until cleanPointPolyData filter is not deleted, the pointer will remain valid.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -333,23 +351,33 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel( vtkMRMLMarkups
...
   cleanedPolyData = cleanPointPolyData->GetOutput();
```
lassoan: This deletes the vtkPolyData object that you created in the previous line (so you should not have created the new vtkPolyData in the previous line) - but anyway, you'll replace all this with a simple pointer.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -333,23 +351,33 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel( vtkMRMLMarkups
...
+  // a lot of operations (vtkOBBTree, computing bounds) seem to fail on vtkPolyData, so we need vtkPoints
+  vtkPoints* cleanedPoints = cleanedPolyData->GetPoints();
+
   vtkSmartPointer< vtkDelaunay3D > delaunay = vtkSmartPointer< vtkDelaunay3D >::New();
   delaunay->SetAlpha( markupsToModelModuleNode->GetDelaunayAlpha() );
   delaunay->AlphaTrisOff();
   delaunay->AlphaLinesOff();
   delaunay->AlphaVertsOff();
+  
+  vtkSmartPointer< vtkMatrix4x4 > transformFromBoundingAxes = vtkSmartPointer< vtkMatrix4x4 >::New();
```
lassoan: transformFromBoundingAxes is a transform, make its name reflect that - for example orientedBoundingBoxToRasTransform.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -333,23 +351,33 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel( vtkMRMLMarkups
...
+  ComputeTransformFromBoundingAxes( cleanedPoints, transformFromBoundingAxes );
+
+  vtkSmartPointer< vtkMatrix4x4 > transformToBoundingAxes = vtkSmartPointer< vtkMatrix4x4 >::New();
+  transformToBoundingAxes->DeepCopy( transformFromBoundingAxes );
```
lassoan: Use vtkMatrix4x4::Invert(const vtkMatrix4x4 *in, vtkMatrix4x4 *out) to do it in one step.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -367,10 +395,14 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel( vtkMRMLMarkups
...
       vtkSmartPointer<vtkRegularPolygonSource> squareSource = vtkSmartPointer<vtkRegularPolygonSource>::New();
-      double extrusionMagnitude = ComputeSurfaceExtrusionAmount( cleanedPoints ); // need to give some depth
       squareSource->SetCenter( 0.0, 0.0, 0.0 );
+      double extrusionMagnitude = ComputeSurfaceExtrusionAmount( smallestBoundingExtentRanges ); // need to give some depth
       squareSource->SetRadius( extrusionMagnitude );
       squareSource->SetNumberOfSides( 4 );
+      double lineAxis[ 3 ] = { 0.0, 0.0, 0.0 }; // temporary values
+      const int LINE_AXIS_INDEX = 0; // The largest (and only meaningful) axis is in the 0th column
+      // the bounding axes are stored in the columns of transformFromBoundingAxes
+      GetNthAxisInMatrix( transformFromBoundingAxes, LINE_AXIS_INDEX, lineAxis );
```
lassoan: GetNthAxisInMatrix is not clear. GetNthColumnOfMatrix would be better. But overall, this helper function just makes things longer. Instead of setting the normal in 4 lines with the help of another function, you could do it all in-place, in one line (it's a long line, so you can break it into 3, but it'll be still shorter and you don't need the additional function):

squareSource->SetNormal( transformFromBoundingAxes->GetElement(0, LINE_AXIS_INDEX), transformFromBoundingAxes->GetElement(1, LINE_AXIS_INDEX), transformFromBoundingAxes->GetElement(2, LINE_AXIS_INDEX) );

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
```diff
@@ -461,214 +500,251 @@ void vtkSlicerMarkupsToModelLogic::UpdateOutputCloseSurfaceModel( vtkMRMLMarkups
...
+  transformFilter->SetTransform( transform );
+  transformFilter->SetInputData( polyDataWithPoints );
+  transformFilter->Update();
+
+  // the extent can be extracted from the output points object (poly data bounds does not work)
+  vtkPoints* transformedPoints = transformFilter->GetPolyDataOutput()->GetPoints();
+  transformedPoints->ComputeBounds();
+  double* extents = transformedPoints->GetBounds(); // { xmin, xmax, ymin, ymax, zmin, zmax }
+
+  double xmin = extents[ 0 ];
```
lassoan: Too verbose. Replace the 4 lines copy-pasted 3x by these 2 lines

for (int axisIndex=0; axisIndex<3; ++axisIndex)
{
  extents[axisIndex]=extents[axisIndex*2+1]-extents[axisIndex*2];
}

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: It's not normal under the current scheme. The output pointer points to a new poly data object. I think this should be changed, but I think it is part of a larger (different) issue around how updates should be handled.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: I understand, I missed this one when going through the code.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: It's not a transform object, but I understand your point. I'll call it a 'matrix'.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: I'll rename to NthColumn. I would prefer to keep the function. I made it in order to compartmentalize how the axes are accessed, and also so the reader doesn't necessarily have to burden him/herself with the lower-level detail.

MarkupsToModel/Logic/vtkSlicerMarkupsToModelLogic.cxx
tavaughan: In my brain-addled state, it took me a while to parse the indices and figure out what the code did. I expanded it a bit more to clarify it.

  for ( int i = 0; i < 3; i++ )
  {
    double axisIMin = extents[ 2 * i ];
    double axisIMax = extents[ 2 * i + 1 ];
    double axisIRange = axisIMax - axisIMin;
    outputExtentRanges[ i ] = axisIRange;
  }

tavaughan:
Thank you for the review Andras, I've addressed (and/or replied) to your comments. I'll wait for your response before merging.

lassoan:
Thank you!

tavaughan:
#125

Issue #54, opened 2015-12-03
Add fiducial matching algorithm


lassoan:
From @mholden8 

There should be a matching algorithm (optional) added to automatically determine the correspondence between fiducial points in the two fiducial lists. Possibly should be made to work when one fiducial list only contains a subset of points.

Add brute force algorithm for fiducial matching.
Refine brute force algorithm to improve performance (perhaps using a branch and bound type algorithm).
Make usable for when there is only a subset of points.
For most practical applications, an polynomial time approximate closed-form solution won't be necessary. A brute force algorithm should be sufficient for 6-10 fiducials.

If there is any ambiguity in the matching, report the best result and give some indicator that there is possible ambiguity in the matching.

tavaughan:
Does this need to work with the "Similarity" and "Warping" transforms?

lassoan:
I think the matching should always use rigid transform. If you select a non-rigid transform you could still perform point matching and outlier rejection are completed using rigid assumption and then use the resulting pointset for non-rigid registration.

If this does not work well then we can simply disable auto-match for non-rigid transforms.

tavaughan:
Might it be worth putting the point correspondence-finding algorithm into its own file/logic? If so, what would be the way to do it? Are there examples in SlicerIGT?

lassoan:
I would put it in the module logic first. If the file grows bigger than a few ten KB then it's definitely better to split out the point matching algorithm into a separate file. There are several C++ modules that have multiple files in the module logic directory.

tavaughan:
There are a few TODO's in the code. Addressing them would make the implementation cleaner so I'm going to do those first.

I should rephrase my earlier question - Are there *preferred* examples in SlicerIGT that I could refer to for multiple logic files? I will likely want to do this later on for MarkupsToModel.

lassoan:
There is nothing special about having multiple files in the logic, just add all files to CMakeLists.txt in the Logic directory. See for example OpenIGTLinkRemote\Logic.

tavaughan:
@mholden8 @lassoan What is the purpose of the InputDataModifiedEvent? In other classes I see it invoked only when the input MRML nodes (markups, transforms, etc) are changed. In this class I see it is also invoked when modes of operation are changed (update manually/automatically, registration rigid/similarity/warping). Is there a reason for this difference?

tavaughan:
Is reflection ever an operation that needs to be taken into consideration?

lassoan:
[...]

PR #131, opened 2017-06-15
Arbitrary point ordering for Fiducial Registration Wizard

tavaughan:
In preparation for #54

### DIFF
--- a/FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.cxx
+++ b/FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.cxx
@@ -1,0 +1,200 @@
+#include "vtkPointDistanceMatrix.h"
+
+#include <vtkDoubleArray.h>
+#include <vtkMath.h>
+#include <vtkObjectFactory.h> //for new() macro
+
+//----------------------------------------------------------------------------
+vtkStandardNewMacro( vtkPointDistanceMatrix );
+
+//------------------------------------------------------------------------------
+vtkPointDistanceMatrix::vtkPointDistanceMatrix()
+{
+  pointList1 = vtkSmartPointer< vtkPoints >::New();
+  pointList2 = vtkSmartPointer< vtkPoints >::New();
+  distanceMatrix = vtkSmartPointer< vtkDoubleArray >::New();
+  UpdateNeededFlag = 0;
+}
+
+//------------------------------------------------------------------------------
+vtkPointDistanceMatrix::~vtkPointDistanceMatrix()
+{
+}
+
+//------------------------------------------------------------------------------
+void vtkPointDistanceMatrix::SetPointList1( vtkPoints* points )
+{
+  if ( points == NULL )
+  {
+    vtkWarningMacro( "Input point list is null. Aborting set operation." );
+    return;
+  }
+
+  pointList1->DeepCopy( points );
+  UpdateNeededFlag = 1;
+}
+
+//------------------------------------------------------------------------------
+void vtkPointDistanceMatrix::SetPointList2( vtkPoints* points )
+{
+  if ( points == NULL )
+  {
+    vtkWarningMacro( "Input point list is null. Aborting set operation." );
+    return;
+  }
+
+  pointList2->DeepCopy( points );
+  UpdateNeededFlag = 1;
+}
+
+//------------------------------------------------------------------------------
+int vtkPointDistanceMatrix::GetPointList1Length()
+{
+  return pointList1->GetNumberOfPoints();
+}
+
+//------------------------------------------------------------------------------
+int vtkPointDistanceMatrix::GetPointList2Length()
+{
+  return pointList2->GetNumberOfPoints();
+}
+
+//------------------------------------------------------------------------------
+double vtkPointDistanceMatrix::GetElement( int pointList1Index, int pointList2Index )
+{
+  int pointList1Length = GetPointList1Length();
+  if ( pointList1Index >= pointList1Length )
+  {
+    vtkWarningMacro( "Point of list 1 is outside the range. Returning 0." )
+    return 0;
+  }
+
+  int pointList2Length = GetPointList2Length();
+  if ( pointList2Index >= pointList2Length )
+  {
+    vtkWarningMacro( "Point of list 2 is outside the range. Returning 0." )
+    return 0;
+  }
+
+  if ( UpdateNeededFlag  )
+  {
+    Update();
+  }
+
+  return distanceMatrix->GetComponent( pointList1Index, pointList2Index );
+}
+
+//------------------------------------------------------------------------------
+double vtkPointDistanceMatrix::GetMinimumElement()
+{
+  if ( distanceMatrix->GetNumberOfTuples() == 0 )
+  {
+    vtkGenericWarningMacro( "Matrix has no contents. Returning 0." )
+    return 0;
+  }
+
+  int pointList1Length = GetPointList1Length();
+  int pointList2Length = GetPointList2Length();
+  double minElement = GetElement( 0, 0 );
+  for ( int pointList1Index = 0; pointList1Index < pointList1Length; pointList1Index++ )
+  {
+    for ( int pointList2Index = 0; pointList2Index < pointList2Length; pointList2Index++ )
+    {
+      double currentElement = GetElement( pointList1Index, pointList2Index );
+      if ( currentElement < minElement )
+      {
+        minElement = currentElement;
+      }
+    }
+  }
+  return minElement;
+}
+
+//------------------------------------------------------------------------------
+void vtkPointDistanceMatrix::Update()
+{
+  int pointList2Length = GetPointList2Length();
+  distanceMatrix->SetNumberOfComponents( pointList2Length );
+  int pointList1Length = GetPointList1Length();
+  distanceMatrix->SetNumberOfTuples( pointList1Length );
+  for ( int pointList1Index = 0; pointList1Index < pointList1Length; pointList1Index++ )
+  {
+    for ( int pointList2Index = 0; pointList2Index < pointList2Length; pointList2Index++ )
+    {
+      double pointInList1[ 3 ] = { 0, 0, 0 }; // temporary values
+      pointList1->GetPoint( pointList1Index, pointInList1 );
+      double pointInList2[ 3 ] = { 0, 0, 0 }; // temporary values
+      pointList2->GetPoint( pointList2Index, pointInList2 );
+      double pointDifference[ 3 ] = { 0, 0, 0 }; // temporary values
+      vtkMath::Subtract( pointInList1, pointInList2, pointDifference );
+      double distance = vtkMath::Norm( pointDifference );
+      distanceMatrix->SetComponent( pointList1Index, pointList2Index, distance );
+    }
+  }
+
+  UpdateNeededFlag = 0;
+}
+
+//------------------------------------------------------------------------------
+void vtkPointDistanceMatrix::ComputeElementWiseDifference( vtkPointDistanceMatrix* matrix1, vtkPointDistanceMatrix* matrix2, vtkDoubleArray* outputArray )
+{
+  if ( matrix1 == NULL )
+  {
+    vtkGenericWarningMacro( "Input first matrix is null. Cannot compute element-wise differences." );
+    return;
+  }
+
+  if ( matrix2 == NULL )
+  {
+    vtkGenericWarningMacro( "Input second matrix is null. Cannot compute element-wise differences." );
+    return;
+  }
+
+  if ( matrix1->GetPointList1Length() != matrix2->GetPointList1Length() )
+  {
+    vtkGenericWarningMacro( "Input matrices have different numbers of points for first list. Cannot compute element-wise differences." );
+    return;
+  }
+
+  if ( matrix1->GetPointList2Length() != matrix2->GetPointList2Length() )
+  {
+    vtkGenericWarningMacro( "Input matrices have different numbers of points for second list. Cannot compute element-wise differences." );
+    return;
+  }
+
+  if ( outputArray == NULL )
+  {
+    vtkGenericWarningMacro( "Output matrix is null. Cannot compute element-wise differences." );
+    return;
+  }
+
+  if ( outputArray->GetNumberOfTuples() > 0 )
+  {
+    vtkGenericWarningMacro( "Output array is not empty. Emptying contents" );
+    outputArray->Reset();
+  }
+
+  int pointList2Length = matrix1->GetPointList2Length();
+  outputArray->SetNumberOfComponents( pointList2Length );
+  int pointList1Length = matrix1->GetPointList1Length();
+  outputArray->SetNumberOfTuples( pointList1Length );
+
+  for ( int pointList1Index = 0; pointList1Index < pointList1Length; pointList1Index++ )
+  {
+    for ( int pointList2Index = 0; pointList2Index < pointList2Length; pointList2Index++ )
+    {
+      double elementFromMatrix1 = matrix1->GetElement( pointList1Index, pointList2Index );
+      double elementFromMatrix2 = matrix2->GetElement( pointList1Index, pointList2Index );
+      double differenceOfElements = elementFromMatrix2 - elementFromMatrix1;
+      outputArray->SetComponent( pointList1Index, pointList2Index, differenceOfElements );
+    }
+  }
+}
+
+void vtkPointDistanceMatrix::PrintSelf(ostream &os, vtkIndent indent)
+{
+  Superclass::PrintSelf(os, indent);
+
+  os << indent << "Point list 1: " << pointList1 << endl;
+  os << indent << "Point list 2: " << pointList2 << endl;
+}
--- a/FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.cxx
+++ b/FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.cxx
@@ -2,7 +2,7 @@
 
 #include <vtkDoubleArray.h>
 #include <vtkMath.h>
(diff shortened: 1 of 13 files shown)


FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.h
```diff
@@ -0,0 +1,49 @@
+#ifndef __vtkPointDistanceMatrix_h
+#define __vtkPointDistanceMatrix_h
+
+#include <vtkDoubleArray.h>
+#include <vtkObject.h>
+#include <vtkPoints.h>
+
+// At its most basic level, this is a Wrapper class for storing a matrix of
```
lassoan: It would be nicer to implement this class as proper VTK filter (takes two vtkPoints objects as input, provides a vtkDoubleArray as output) - if you are interested in learning how to implement such filters.

FiducialRegistrationWizard/Logic/vtkSlicerFiducialRegistrationWizardLogic.cxx
```diff
@@ -254,32 +254,48 @@ bool vtkSlicerFiducialRegistrationWizardLogic::UpdateCalibration( vtkMRMLNode* n
...
+  MarkupsFiducialNodeToVTKPoints( toMarkupsFiducialNode, toPointsUnordered );
+
+  // Determine the order of points and store an "ordered" version of the "To" list
+  vtkSmartPointer< vtkPoints > toPointsOrdered = NULL; // temporary value
+  int inputFormat = fiducialRegistrationWizardNode->GetInputFormat();
+  if ( inputFormat == vtkMRMLFiducialRegistrationWizardNode::INPUT_FORMAT_ORDERED_PAIRS )
+  {
+    toPointsOrdered = toPointsUnordered;
+  }
+  else if ( inputFormat == vtkMRMLFiducialRegistrationWizardNode::INPUT_FORMAT_UNORDERED_PAIRS )
```
lassoan: Often we accidentally record two points or forget to record a point. Can you deal with point sets that have different lengths?

FiducialRegistrationWizard/Logic/vtkSlicerFiducialRegistrationWizardLogic.cxx
```diff
@@ -442,6 +458,203 @@ bool vtkSlicerFiducialRegistrationWizardLogic::CheckCollinear( vtkPoints* points
...
+    vtkGenericWarningMacro( "Input array is null. Returning 0." );
+    return 0;
+  }
+
+  double sumOfSquaredElements = 0;
+  int numberOfTuples = array->GetNumberOfTuples();
+  int numberOfComponents = array->GetNumberOfComponents();
+  for ( int tupleIndex = 0; tupleIndex < numberOfTuples; tupleIndex++ )
+  {
+    for ( int componentIndex = 0; componentIndex < numberOfComponents; componentIndex++ )
```
lassoan: what are the components in the array?

FiducialRegistrationWizard/Logic/vtkSlicerFiducialRegistrationWizardLogic.cxx
```diff
@@ -442,6 +458,203 @@ bool vtkSlicerFiducialRegistrationWizardLogic::CheckCollinear( vtkPoints* points
...
+  outputPermutationsArray->SetNumberOfTuples( numberOfPossiblePermutations );
+
+  // The functions will work in place on a vtkIntArray in order to speed up operations,
+  // avoid allocation and deallocation.
+  vtkSmartPointer< vtkIntArray > array = vtkSmartPointer< vtkIntArray >::New();
+  array->SetNumberOfComponents( numberOfIndicesToPermute );
+  array->SetNumberOfTuples( 1 );
+  for ( int index = 0; index < numberOfIndicesToPermute; index++ )
+  {
+    array->SetComponent( 0, index, index ); // the the index'th value to index
```
lassoan: the the

FiducialRegistrationWizard/MRML/vtkMRMLFiducialRegistrationWizardNode.cxx
```diff
@@ -319,6 +320,56 @@ int vtkMRMLFiducialRegistrationWizardNode::UpdateModeFromString( std::string upd
...
+}
+
+//------------------------------------------------------------------------------
+std::string vtkMRMLFiducialRegistrationWizardNode::InputFormatAsString( int inputFormat )
+{
+  switch ( inputFormat )
+  {
+    case INPUT_FORMAT_ORDERED_PAIRS:
+    {
+      return "Ordered Pairs";
```
lassoan: This string should be for machine reading (to write to XML). The best is to store strings that will be displayed to users in the .ui file. If you store the displayable string in the source file then use the tr() macro to allow translation.

FiducialRegistrationWizard/MRML/vtkMRMLFiducialRegistrationWizardNode.cxx
```diff
@@ -319,6 +320,56 @@ int vtkMRMLFiducialRegistrationWizardNode::UpdateModeFromString( std::string upd
...
+    default:
+    {
+      vtkGenericWarningMacro( "Unrecognized value for InputFormat " << inputFormat << ". Returning \"Unknown\"" );
+      return "Unknown";
+    }
+  }
+}
+
+//------------------------------------------------------------------------------
+int vtkMRMLFiducialRegistrationWizardNode::InputFormatFromString( std::string inputFormatAsString )
```
lassoan: InputFormat is not a very expressive name. Something like PointMatchingMethod would be better.

FiducialRegistrationWizard/Logic/vtkSlicerFiducialRegistrationWizardLogic.h
```diff
@@ -84,9 +84,16 @@ class VTK_SLICER_FIDUCIALREGISTRATIONWIZARD_MODULE_LOGIC_EXPORT vtkSlicerFiducia
   virtual void OnMRMLSceneNodeRemoved(vtkMRMLNode* node);
 
 private:
+
   vtkSlicerFiducialRegistrationWizardLogic(const vtkSlicerFiducialRegistrationWizardLogic&); // Not implemented
   void operator=(const vtkSlicerFiducialRegistrationWizardLogic&);               // Not implemented
 
+  static double SumOfSquaredElementsInArray( vtkDoubleArray* array );
```
lassoan: For me, it's not obvious how all this works. Adding a one-sentence description for each method would help a lot.

tavaughan:
Oops, it looks like my commit wound up in this pull request. Please hold off on merging until I do a bit more cleanup. :)

tavaughan:
I see you just provided feedback! Sorry you had to sift through somewhat messy work-in-progress code... The feedback is still helpful though, so thank you!

FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.cxx
```diff
@@ -0,0 +1,200 @@
...
+  if ( UpdateNeededFlag  )
+  {
+    Update();
+  }
+
+  return distanceMatrix->GetComponent( pointList1Index, pointList2Index );
+}
+
+//------------------------------------------------------------------------------
+double vtkPointDistanceMatrix::GetMinimumElement()
```
tavaughan: That's a good point. I've made the cahnge

FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.cxx
```diff
@@ -0,0 +1,200 @@
...
+{
+  int pointList2Length = GetPointList2Length();
+  distanceMatrix->SetNumberOfComponents( pointList2Length );
+  int pointList1Length = GetPointList1Length();
+  distanceMatrix->SetNumberOfTuples( pointList1Length );
+  for ( int pointList1Index = 0; pointList1Index < pointList1Length; pointList1Index++ )
+  {
+    for ( int pointList2Index = 0; pointList2Index < pointList2Length; pointList2Index++ )
+    {
+      double pointInList1[ 3 ] = { 0, 0, 0 }; // temporary values
```
tavaughan: I wrote it out on multiple lines to make it more human-readable. But if the code works then I will reduce its length by using the this->PointList1->GetPoint(pointList1Index) and the Distance2BetweenPoints() functions

FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.cxx
```diff
@@ -0,0 +1,200 @@
...
+//----------------------------------------------------------------------------
+vtkStandardNewMacro( vtkPointDistanceMatrix );
+
+//------------------------------------------------------------------------------
+vtkPointDistanceMatrix::vtkPointDistanceMatrix()
+{
+  pointList1 = vtkSmartPointer< vtkPoints >::New();
+  pointList2 = vtkSmartPointer< vtkPoints >::New();
+  distanceMatrix = vtkSmartPointer< vtkDoubleArray >::New();
+  UpdateNeededFlag = 0;
```
tavaughan: This sounds like a really good idea. I noticed that in other classes (vtkLookupTable, vtkImageData, etc...) that the variable is declared **not** as a pointer. Is there a particular reason you suggest declaring it as a pointer, or should I just follow the examples in vtk?

lassoan: If it works as a simple variable and not a pointer then it's even better, as it's simpler.

FiducialRegistrationWizard/Logic/vtkSlicerFiducialRegistrationWizardLogic.cxx
tavaughan: It shouldn't require many changes.

FiducialRegistrationWizard/MRML/vtkMRMLFiducialRegistrationWizardNode.cxx
tavaughan: I don't understand, and I don't see tr() used in many places in SlicerIGT. What is the problem that it is supposed to solve? Is it the space in "Ordered Pairs" when writing to XML...?

Where should I use the tr() function?

FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.h
tavaughan: I think I'd rather leave it as-is for now and change if and when needed.

tavaughan:
* not ready for merging yet, I still need to add support for unequally-sized inputs

tavaughan:
*still not ready for merging* - though Andras if you want to look at vtkCombinatoricGenerator, that class shouldn't change much (if at all) between now and the merge-ready version

tavaughan:
#54 is addressed. Code review would be welcome, I'll also take another look over this work in the next couple of days.

There are already a few outstanding issues (either unfinished from this work, or that I otherwise noticed), might be best to create separate tickets:
- Currently no detection when point matching is ambiguous (two or more equally-suited solutions)
- Point matching does not yet handle uniform scale.
- Long error messages to "Status" label cause the Slicer GUI to resize.
- Smarter + faster point-matching would be a nice feature to implement..
- For larger point sets (>~6), a few representative points could be chosen and used to come up with an initial guess/registration. Then other points can be matched based on initial guess/registration.

I hope smaller issues like these can be worked out individually.

FiducialRegistrationWizard/Logic/vtkPointMatcher.cxx
```diff
@@ -0,0 +1,368 @@
...
+  os << indent << "IsMatchingWithinTolerance: " << this->IsMatchingWithinTolerance() << std::endl;
+  os << indent << "UpdateNeeded: " << this->UpdateNeeded() << std::endl;
+}
+
+//------------------------------------------------------------------------------
+// INPUT MUTATORS
+//------------------------------------------------------------------------------
+
+//------------------------------------------------------------------------------
+void vtkPointMatcher::SetInputPointList1( vtkPoints* points )
```
lassoan: In general, in VTK classes Set(Object) functions do not make a deep copy.
See for example in vtkRenderView:

header

    virtual void SetTransform(vtkAbstractTransform* transform);
    vtkGetObjectMacro(Transform, vtkAbstractTransform);

implementation

    vtkCxxSetObjectMacro(vtkRenderView, Transform, vtkAbstractTransform);

You can use vtkSetObjectBodyMacro if you need to do some custom operations in addition to just storing a reference to the object.

FiducialRegistrationWizard/Logic/vtkPointMatcher.cxx
```diff
@@ -0,0 +1,368 @@
...
+  this->TolerableRootMeanSquareDistanceErrorMm = errorMm;
+  this->OnInputChanged();
+}
+
+//------------------------------------------------------------------------------
+void vtkPointMatcher::OnInputChanged()
+{
+  // mean distance error has not been computed yet. Set to maximum possible value for now
+  this->ComputedRootMeanSquareDistanceErrorMm = RESET_VALUE_COMPUTED_ROOT_MEAN_DISTANCE_ERROR;
+  this->InputChangedTime.Modified();
```
lassoan: why don't you just call this->Modified()?

FiducialRegistrationWizard/Logic/vtkPointMatcher.cxx
```diff
@@ -0,0 +1,368 @@
...
+  double meanOfSquaredDistanceErrors = sumOfSquaredDistanceErrors / numberOfDistances;
+  double rootMeanSquareistanceErrors = sqrt( meanOfSquaredDistanceErrors );
+
+  return rootMeanSquareistanceErrors;
+}
+
+//------------------------------------------------------------------------------
+bool vtkPointMatcher::UpdateNeeded()
+{
+  return ( this->InputChangedTime > this->OutputChangedTime );
```
lassoan: If you use standard VTK get/set methods, those update ModifiedTime of the object, so you don't need to maintain a separate InputChangedTime variable.

FiducialRegistrationWizard/MRML/vtkMRMLFiducialRegistrationWizardNode.cxx
```diff
@@ -49,8 +55,9 @@ vtkMRMLFiducialRegistrationWizardNode::vtkMRMLFiducialRegistrationWizardNode()
   this->AddNodeReferenceRole( FROM_FIDUCIAL_LIST_REFERENCE_ROLE, NULL, fiducialListEvents.GetPointer() );
   this->AddNodeReferenceRole( TO_FIDUCIAL_LIST_REFERENCE_ROLE, NULL, fiducialListEvents.GetPointer() );
   this->AddNodeReferenceRole( OUTPUT_TRANSFORM_REFERENCE_ROLE );
-  this->RegistrationMode = "Rigid";
-  this->UpdateMode = "Automatic";
+  this->RegistrationMode = REGISTRATION_MODE_RIGID;
+  this->UpdateMode = UPDATE_MODE_MANUAL;
```
lassoan: this->UpdateMode = "Automatic" => this->UpdateMode = UPDATE_MODE_MANUAL

Why is the change?

FiducialRegistrationWizard/MRML/vtkMRMLFiducialRegistrationWizardNode.cxx
```diff
@@ -253,6 +289,107 @@ void vtkMRMLFiducialRegistrationWizardNode::SetUpdateMode( std::string newUpdate
...
+{
+  switch ( method )
+  {
+    case POINT_MATCHING_METHOD_INPUT_ORDER:
+    {
+      return "Input Order";
+    }
+    case POINT_MATCHING_METHOD_COMPUTED:
+    {
+      return "Computed";
```
lassoan: For me, it is not clear what you mean by "Input order". Also, there may be multiple computation methods in the future. So, the options could be:
- original order (not very good either, so ask a few people if they understand what we mean by this)
- automatic reorder

FiducialRegistrationWizard/MRML/vtkMRMLFiducialRegistrationWizardNode.cxx
tavaughan: Possibly a typo on my part. Will look into it.

FiducialRegistrationWizard/MRML/vtkMRMLFiducialRegistrationWizardNode.cxx
tavaughan: Or even something as simple as "Manual order" vs "Automatic order". I'll try to gather a few ideas.

lassoan: I like manual / automatic ordering

FiducialRegistrationWizard/Logic/vtkPointDistanceMatrix.cxx
```diff
@@ -0,0 +1,226 @@
...
+    vtkGenericWarningMacro( "Output array is not empty. Emptying contents" );
+    outputArray->Reset();
+  }
+
+  int pointList2Length = matrix1->GetPointList2Length();
+  outputArray->SetNumberOfComponents( pointList2Length );
+  int pointList1Length = matrix1->GetPointList1Length();
+  outputArray->SetNumberOfTuples( pointList1Length );
+
+  for ( int pointList1Index = 0; pointList1Index < pointList1Length; pointList1Index++ )
```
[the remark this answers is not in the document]
tavaughan: Not sure I see the connection. There are two point list indices because I want to preserve information about which point-to-point distance is which.

tavaughan:
Verified that this compiles on Windows. Merging.

PR #133, opened 2017-06-22
BUG: MarkupsToModel - should not be possible to deselect radio button.

### LINKED ISSUE #132: MarkupsToModel Radio buttons can be deselected

tavaughan:
Fix #132


lassoan:
All items were in a grid layout to make them all nicely aligned. If you put widgets in separate groupboxes then the layout will be messy.

Could you please restore the original layout and instead define QButtonGroup for the radiobuttons in the setup method of the module widget?

tavaughan:
Is it not still in a grid layout and aligned? This seems consistent with how the other radio buttons are set up.

lassoan:
Thank you @tavaughan !

PR #142, opened 2017-10-13
General updates for transform fusion

tavaughan:
- Added transform concatentation/inversion support with ComputeFullTransform mode
- Fixed automatic updating for transform fusion
- Don't store result in general transform, otherwise changes anywhere will get reflected


tavaughan:
Verified that it compiles on Windows. Merging.

PR #157, opened 2017-11-07
BUG: Fix bug in OpenIGTLink Remote causing Slicer to crash

Sunderlandkyl:
Changing the query data type when the list has been populated with items will cause the items in the list to become empty.
Selecting one of these empty items and clicking "Get selected items" will cause Slicer to crash.

Fix is two-fold:
- After query data type is changed, set the number of rows to zero, so that there are no empty items in the list.
- When attempting to get a selected item in querySelectedItem(), check to see if the item exists. If it does not, print an error message and skip the currently selected item.


lassoan:
Good catch, thank you!

PR #167, opened 2018-02-02
EHN: Remove the dependency on OpenIGTLink library.

leochan2009:


leochan2009:
It solves a problem. As the OpenIGTLink library is removed from Slicer, so OpenIGTLinkRemote won't find the library anymore. That is the reason why build of SlicerIGT failed on the server.
http://slicer.cdash.org/buildSummary.php?buildid=1192463
The configuration file of OpenIGTLinkIF will include OpenIGTLink, OpenIGTLinkIO libraries.

lassoan:
Ok. Thank you for the fix!

PR #174, opened 2018-03-19
ENH: Migrate PlusRemote and OpenIGTLinkRemote to SlicerOpenIGTLink

Sunderlandkyl:
- Migrate PlusRemote and OpenIGTLinkRemote to SlicerOpenIGTLink (https://github.com/openigtlink/SlicerOpenIGTLink/pull/19)

PR #175, opened 2018-03-20
Removing OpenIGTLinkRemote as it has migrated to SlicerOpenIGTLink extension

adamrankin:
Remove entry in extension CMakeList

Remove folder containing module


lassoan:
I think this PR https://github.com/SlicerIGT/SlicerIGT/pull/174 has taken care of this already. I've just merged it now. Let me know if something is missed.

PR #177, opened 2018-04-04
Guidelet fixes for 4.9

xTdub:


Guidelet/GuideletLib/Guidelet.py
```diff
@@ -545,7 +545,7 @@ def removeConnectorObservers(self):
 
   def addConnectorObservers(self):
     # Force initial update
-    if self.connectorNode.GetState() == slicer.vtkMRMLIGTLConnectorNode.STATE_CONNECTED:
+    if self.connectorNode.GetState() == slicer.vtkMRMLIGTLConnectorNode.StateConnected:
```
ungi: This won't work with Slicer 4.8.1. Due to a controversial decision we use the same SlicerIGT version for latest stable and nightly Slicer builds. I don't know if there is a way to check Slicer version in the code, but you could check vtk.VTK_MAJOR_VERSION. It should be 7 for Slicer 4.8 and 9 for Slicer 4.9. You could use the old or new constant based on VTK version.

ungi: @lassoan is there a more appropriate way to check Slicer version?

lassoan: SlicerIGT is built from a different branch for Slicer-4.8 (master-48), so in the master branch we don't need to maintain backward compatibility with Slicer-4.8.

PR #180, opened 2018-06-14
Fix configuration errors when BUILD_TESTING is OFF

auneri:
... by preventing BreachWarning from trying to set properties of non-existent targets.


lassoan:
Thank you!

PR #184, opened 2018-11-05
Dynamic guidelet layouts

cpinter:


cpinter:
@ungi I'd appreciate if you could test at least one or two existing guidelets with these changes. (The second commit will cause errors at exiting, but as there are errors currently when exiting it's not a big difference, just different errors)

@lassoan Please check out the second commit about the cleanup function. Thanks!

cpinter:
Thanks, Tamas!

@lassoan Can you check out the second commit about the cleanup function? Thanks!

lassoan:
Looks good to me, thank you!

Issue #188, opened 2019-01-13
WatchdogSelfTest failing when functionality works


adamrankin:
I thought I had fixed the test in 3e619113663edfc6d547977aa7f744e7956f5d07 but obviously not.

http://slicer.cdash.org/viewTest.php?onlyfailed&buildid=1486792

I believe it's the testing environment that does not allow the qApp to update normally due to being in the python script.

Unfortunately, with a failing test, downstream extensions are not built or uploaded to extension manager.

PR #189, opened 2019-01-14
disable test

adamrankin:
re #188 disable test to enable downstream extensions to build until solution can be found


adamrankin:
@lassoan please confirm if it is ok to temporarily disable the test until it's fixed

lassoan:
Yes, it's fine to temporarily disable the test

PR #190, opened 2019-03-12
Update code to use new markups naming

adamrankin:
UNTESTED


adamrankin:
This just fixes the build, I have not tested the affected modules for proper behaviour.

lassoan:
Thank you for the suggestions, SlicerIGT now builds with 4.10 and master branch of Slicer.

PR #191, opened 2019-03-22
ENH: Add node references in the registration modules

cpinter:
This is related to the recent changes in Slicer, SlicerExecutionModel, and BRAINSTools, the goal of which is to update registration modules to have the output transforms reference the inputs (images/fiducial lists/models, fixed and moving separately).


cpinter:
@ungi @lassoan If there are no objections I'll merge this. Thanks!

ungi:
Thanks. It doesn't change anything from the perspective of module users, right?

cpinter:
No. It just adds node references

PR #192, opened 2019-06-24
Fixed Python 3 absolute imports and unicode function. 

jlaframboise:
I added absolute imports. I changed the unicode() function to the str.encode() string method for Python 3. I removed a condition that checked unicode() in addition to str() because they are now the same in Python 3.

### DIFF
--- a/Guidelet/GuideletLib/Guidelet.py
+++ b/Guidelet/GuideletLib/Guidelet.py
@@ -41,17 +41,17 @@ class Guidelet(object):
     else:
       logging.debug("Timeout. Command Id: {0}".format(commandId))
 
-    # Guidelet layout name definitions
-    VIEW_ULTRASOUND = unicode("Ultrasound")
-    VIEW_ULTRASOUND_3D = unicode("Ultrasound + 3D")
-    VIEW_3D_ULTRASOUND = unicode("3D + Ultrasound")
-    VIEW_ULTRASOUND_CAM_3D = unicode("Ultrasound + Webcam + 3D")
-    VIEW_ULTRASOUND_DUAL_3D = unicode("Ultrasound + Dual 3D")
-    VIEW_3D = unicode("3D")
-    VIEW_DUAL_3D = unicode("Dual 3D")
-    VIEW_TRIPLE_3D = unicode("Triple 3D")
-    VIEW_TRIPLE_3D_PARALLEL = unicode("Triple 3D Parallel")
-    VIEW_QUAD_3D = unicode("Quad 3D")
+  # Guidelet layout name definitions
+  VIEW_ULTRASOUND = unicode("Ultrasound")
+  VIEW_ULTRASOUND_3D = unicode("Ultrasound + 3D")
+  VIEW_3D_ULTRASOUND = unicode("3D + Ultrasound")
+  VIEW_ULTRASOUND_CAM_3D = unicode("Ultrasound + Webcam + 3D")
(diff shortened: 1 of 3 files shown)


Guidelet/GuideletLib/Guidelet.py
```diff
@@ -38,16 +38,16 @@ def onGenericCommandResponseReceived(commandId, responseNode):
...
-  VIEW_ULTRASOUND_3D = unicode("Ultrasound + 3D")
-  VIEW_3D_ULTRASOUND = unicode("3D + Ultrasound")
-  VIEW_ULTRASOUND_CAM_3D = unicode("Ultrasound + Webcam + 3D")
-  VIEW_ULTRASOUND_DUAL_3D = unicode("Ultrasound + Dual 3D")
-  VIEW_3D = unicode("3D")
-  VIEW_DUAL_3D = unicode("Dual 3D")
-  VIEW_TRIPLE_3D = unicode("Triple 3D")
-  VIEW_TRIPLE_3D_PARALLEL = unicode("Triple 3D Parallel")
-  VIEW_QUAD_3D = unicode("Quad 3D")
+  VIEW_ULTRASOUND = ("Ultrasound").encode
```
Sunderlandkyl: I think you mean ("").encode()

Guidelet/GuideletLib/Guidelet.py
```diff
@@ -38,16 +38,16 @@ def onGenericCommandResponseReceived(commandId, responseNode):
...
+  VIEW_ULTRASOUND_3D = ("Ultrasound + 3D").encode
+  VIEW_3D_ULTRASOUND = ("3D + Ultrasound").encode
+  VIEW_ULTRASOUND_CAM_3D = ("Ultrasound + Webcam + 3D").encode
+  VIEW_ULTRASOUND_DUAL_3D = ("Ultrasound + Dual 3D").encode
+  VIEW_3D = ("3D").encode
+  VIEW_DUAL_3D = ("Dual 3D").encode
+  VIEW_TRIPLE_3D = ("Triple 3D").encode
+  VIEW_TRIPLE_3D_PARALLEL = ("Triple 3D Parallel").encode
+  VIEW_QUAD_3D = ("Quad 3D").encode
```
Sunderlandkyl: Same for the rest of these.

Guidelet/GuideletLib/Guidelet.py
```diff
@@ -221,7 +221,7 @@ def setupAdditionalPanel(self):
     pass
 
   def registerLayout(self, layoutName, layoutId, layoutXmlDescription, layoutSelectCallback=None):
-    if (type(layoutName) != str and type(layoutName) != unicode) or len(layoutName) < 1:
+    if (type(layoutName) != str) or len(layoutName) < 1:
```
Sunderlandkyl: This code has to work in the Stable version of Slicer as well. type(layoutName) will be 'unicode' in Slicer 4.10.2.

If there isn't a way to do type checking that works on both platforms, you can check the Slicer version before determining the type using:
```
if slicer.app.majorVersion >= 5 or (slicer.app.majorVersion >= 4 and slicer.app.minorVersion >= 11):
```

Guidelet/GuideletLib/Guidelet.py
jlaframboise: Good catch. Although I think instead of using encode I will alias str as unicode for Python 3.

PR #194, opened 2019-07-05
Add cast to list around map for Python 3 compatibility

zacbaum:
Map (along with several other functions) no longer return lists in Python 3, this casts the iterable into a list to allow the rest of the function to execute as normal.


lassoan:
Thank you!

PR #196, opened 2020-04-08
Move MetafileImporter module here from Sequences extension for Slicer-4.11

lassoan:
Sequences extension is moved to Slicer-4.11, but sequence metafile importer does not belong to Slicer core, therefore it is moved here.


jamesobutler:
@lassoan Why does Sequence Metafile importer not belong in Slicer core with the other sequence work? I currently use the Sequences extension and also specifically use this importer. I haven’t required any of the rest of SlicerIGT before.

lassoan:
Sequence metafile (tracked image data) format is specific to SlicerIGT. Moving it to Slicer core would make the core slightly larger and make the feature harder to modify in stable versions. Maybe we could move it to SlicerIGSIO extension, that is much smaller than SlicerIGT.

jamesobutler:
Yeah I think SlicerIGSIO would make sense as there is a SequenceIO module already there too. We use it for reading some PLUS captured tracked meta image files.

jamesobutler:
Also I guess SlicerIGSIO and other extensions will need to update their dependencies once Sequences gets merged into Slicer core.

lassoan:
We may keep an empty Sequences extension around for a while to avoid breaking builds (it could just contain a single scripted module that logs a warning that sequences is deprecated, maybe also create an alias of slicer.modules.sequencebrowser to slicer.modules.sequences).

jamesobutler:
@lassoan did you not want to add it to SlicerIGSIO?

lassoan:
Yes, I'll add metafileimporter to SlicerIGSIO.

lassoan:
Done - https://github.com/IGSIO/SlicerIGSIO/commit/351741a58d173ce81b084ccfedf2628fccaaddc7

PR #197, opened 2020-04-22
Fix customapp build and deployment

cpinter:
The first commit fixes customapp build. The removed CMake commands were not present in any other module, so it seems safe.

The second commit fixes resource deployment (sound files for BreachWarning and Watchdog) in a way that it works both when using it as Slicer extension and in custom app. Packaging has been tested as well.
This commit also renames the folder (and the few occurrences) of ToolWatchdog to Watchdog for consistency.


BreachWarning/CMakeLists.txt
```diff
@@ -63,22 +63,9 @@ endif()
...
 configure_file(
   ${CMAKE_CURRENT_SOURCE_DIR}/Resources/alarm.wav
-  ${SlicerIGT_BINARY_DIR}/${Slicer_QTLOADABLEMODULES_SHARE_DIR}/${MODULE_NAME}/alarm.wav
+  ${CMAKE_BINARY_DIR}/${Slicer_QTLOADABLEMODULES_SHARE_DIR}/${MODULE_NAME}/alarm.wav
   COPYONLY)
 
 install(
   FILES ${CMAKE_CURRENT_SOURCE_DIR}/Resources/alarm.wav
   DESTINATION ${Slicer_INSTALL_QTLOADABLEMODULES_SHARE_DIR}/${MODULE_NAME} COMPONENT Runtime)
-
```
lassoan, line 72: Have you removed these intentionally?

cpinter: You mean the last empty lines? No. I'll put them back.

cpinter: Strange... When I opened this file, the last blank line is there.

lassoan: I mean if removing the `SET_TARGET_PROPERTIES(...PROPERTIES FOLDER ...)` lines (that put modules in subfolders in the generated Visual Studio solution) was intentional.

cpinter: That broke the custom app build. I assumed it was safe to remove, because these commands were not present in any other module.

lassoan: Very, strange, how did it break things? Folders are used for every Slicer core module. It is missed in extensions because it was not included in the extension template. I'll merge this in the meantime to not hold up the other changes.

cpinter: Thanks @lassoan! When I first made the custom app in September 2019 the build failed (not sure about the error anymore) and removing these fixed the problem. I can try to re-add it and see if it builds now. Should I?

PR #198, opened 2020-05-19
BUG: Fix error on startup in Guidelet

cpinter:
Creating two layouts for the same frame caused the error "QLayout: Attempting to add QLayout "" to QFrame "", which already has a layout" to be thrown when starting a Guidelet. This is a problem if one tries to rely on the error icon for actual errors, and makes reading logs more difficult.


cpinter:
Thanks @lassoan!

PR #200, opened 2020-11-26
Update TextureModel.py

ezgimercan:
Per [this ](https://discourse.slicer.org/t/mutli-texture-rendering-with-obj-mlt-multiple-jpgs/14773/2?u=ezgimercan ) suggestion on discourse, multi-texture model import added to TextureModel module. It loads the model with texture using vtkOBJImporter class and saves the actor and texture in Slicer temp folder using vtkSingleVTPExporter class. Then reads VTP and PNG into scene and applies texture.


fbordignon:
Very cool solution, isn't something we would want to merge? If it is missing something I can work on it. Thanks!

PR #205, opened 2021-12-02
ENH: Add Watchdog option to only monitor ModifiedEvent

Sunderlandkyl:
Previously, it was impossible to tell if a transform was up-to-date if any parent transform were also being modified.
This was because TransformModifiedEvent is invoked when any parent transform is changed.

This commit adds the vtkMRMLWatchdogNode::OnlyWatchModifiedEvents option, which will only update the watchdog when vtkCommand::ModifiedEvent are observed.


lassoan (commented):
Thank you, it looks good, except I would rename the "onlyWatchModifiedEvents" option to be more future-compatible. The issue with "onlyXYZ" type of flags that we may be incompatible with options that we add in the future (when we want to observe not just Modified events, but we still want to exclude TransformModified events). Instead, if we called the flag "watchTransformModifiedEvents" then that will remain meaningful regardless of what additional features will be added.

Sunderlandkyl:
Thanks for the review! I've updated the option.

PR #206, opened 2022-04-10
COMP: Mark overriding functions with override specifier

markasselin:
Adding override specifiers, as these are considered good practice for modern C++ code. Without the specifiers SlicerIGT emits >800 unnecessary warnings on macOS.


Sunderlandkyl:
Great, thanks for doing this!

PR #207, opened 2022-04-14
ENH: Update pivot calibration algorithm to use IGSIO

Sunderlandkyl:
This commit changes the internal logic of vtkSlicerPivotCalibrationLogic to use the pivot and spin calibration algorithms from IGSIO.

This also provides access to the option to enable automatic calibration from Python. During automatic calibration, the buffer will be filled as the tool moves. If invalid tool motion is detected due to high error in recent poses, then the whole input buffer will be discarded. Once the required number of poses and error threshold is reached, vtkSlicerPivotCalibrationLogic will invoke (Pivot/Spin)CalibrationCompleteEvent. These options are currently not accessible through the GUI.


lassoan (commented):
Thank you, it looks good to me.

PR #208, opened 2022-05-25
COMP: Fix build in Slicer custom apps

cpinter:
See discussion https://discourse.slicer.org/t/extension-dependencies-in-custom-apps/23552/8

Co-authored-by: Kyle Sunderland <krs2@queensu.ca>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,13 @@ set(EXTENSION_DEPENDS SlicerIGSIO)
 # Extension dependencies
 find_package(Slicer REQUIRED)
 include(${Slicer_USE_FILE})
-find_package(SlicerIGSIO REQUIRED)
+
+if (NOT DEFINED Slicer_EXTENSION_SOURCE_DIRS)
+  find_package(SlicerIGSIO REQUIRED) 
+else()
+  # Allow usage if dependent extension is bundled
+  find_package(IGSIO REQUIRED) 
+endif()
 
 #-----------------------------------------------------------------------------
 # Extension modules


CMakeLists.txt
```diff
@@ -20,7 +20,13 @@ set(EXTENSION_DEPENDS SlicerIGSIO)
 # Extension dependencies
 find_package(Slicer REQUIRED)
 include(${Slicer_USE_FILE})
-find_package(SlicerIGSIO REQUIRED)
+
+if (NOT DEFINED Slicer_EXTENSION_SOURCE_DIRS)
+  find_package(SlicerIGSIO REQUIRED) 
+else()
+  # Allow usage in custom applications
```
lassoan, line 27: I'm not sure if this comment is accurate. The difference should not be due to bundling the extension into Slicer but because there was a reorganization of libraries between packages in recent Slicer versions.

lassoan (commented):
It looks good but please double-check with @Sunderlandkyl if the difference is really due to bundling the extension.

CMakeLists.txt
cpinter: I can change this to
`# Allow usage for bundled extensions`
I wasn't sure what were bundled extensions, but based on what I found in the code, such as 
https://github.com/Slicer/Slicer/blob/96ee24860f7e9cb7364f5e72872e6d0a0cbd097f/CMake/SlicerExtensionCPackBundleFixup.cmake.in#L230-L234
it seems that it is primarily for custom apps. Up to you, I don't mind either way, just wanted to document a bit why the difference.

cpinter: Maybe better I guess
`# Allow usage if dependent extension is bundled`

cpinter:
I updated the PR to have the comment `# Allow usage if dependent extension is bundled` and merged it if you don't mind. I can send a follow-up fix if needed. Thanks

CMakeLists.txt
```diff
@@ -20,7 +20,13 @@ set(EXTENSION_DEPENDS SlicerIGSIO)
 # Extension dependencies
 find_package(Slicer REQUIRED)
 include(${Slicer_USE_FILE})
-find_package(SlicerIGSIO REQUIRED)
+
+if (NOT DEFINED Slicer_EXTENSION_SOURCE_DIRS)
```
lassoan, line 24: What happens if SlicerIGSIO is bundled but SlicerIGT is not? Does it still work?
Maybe it would be better to try `find_package(IGSIO)` and if it is not found then try finding `find_package(SlicerIGSIO REQUIRED)`?

PR #213, opened 2023-07-08
ENH: Add QTableWidget to qSlicerPathExplorerModuleWidget

RafaelPalomar:
This change is related to Slicer/Slicer#6909, which fixes the inclusion of generated (and private) Qt `ui_*` files; this indirectly included some Qt headers, which naturally would be missing. This commit includes the `QTableWidget` that would be required upon merging of Slicer/Slicer#6909. This change works also with Slicer/Slicer#6909 not merged.


Sunderlandkyl:
Thanks!

PR #221, opened 2024-07-31
ENH: Enhance sequence playback control

dalbenzioG:


SequenceReplay/SequenceReplay.py
```diff
@@ -255,95 +256,103 @@ def setParameterNode(self, inputParameterNode: Optional[SequenceReplayParameterN
...
         slicer.modules.sequences.toolBar().setActiveBrowserNode(sequenceBrowserNode)
-        
-        # Iterate through all sequence browser nodes. Stop replay in all except the clicked one. Start play in the clicked one.
-        for node in slicer.util.getNodesByClass("vtkMRMLSequenceBrowserNode"):
-            if node.GetID() == sequenceBrowserNodeId:
-                node.SetPlaybackActive(True)
+
+        # Toggle playback state
+        if sequenceBrowserNodeId in self.playbackState:  # If the node is currently playing
+            if self.playbackState[sequenceBrowserNodeId]:
```
lassoan, line 292: Make all string that is displayed to the user translatable. Something like this (not tested):

```
            if playback:
                buttonText = _("Play {name}").format(name=sequenceBrowserNode.GetName())
            else:
                buttonText = _("Stop {name}").format(name=sequenceBrowserNode.GetName())
            self.buttonState[sequenceBrowserNodeId].setText(buttonText)
```

SequenceReplay/SequenceReplay.py
```diff
@@ -255,95 +256,103 @@ def setParameterNode(self, inputParameterNode: Optional[SequenceReplayParameterN
...
-            widget.setParent(None)
-            widget.deleteLater()
-            
-        # Create a button for each node
+        self._clearLayout(sequenceBrowsersLayout)
+
+        sequenceBrowserNodes = slicer.util.getNodesByClass("vtkMRMLSequenceBrowserNode")
         for node in sequenceBrowserNodes:
-            button = qt.QPushButton(node.GetName())
+            button = qt.QPushButton("Play " + node.GetName())
```
lassoan, line 325: It would be better if all the button creation/deletion/updates were done here, triggered by MRML node events.

SequenceReplay/SequenceReplay.py
lassoan, line 292: `node.SetPlaybackActive(True)` is all that is needed in response to GUI actions. The buttons will be updated in response to MRML node updates (that you notice because there is an observer on each browser node). It is important not to update the buttons state and text in the GUI event handler, because the browser nodes can be changed for many other reasons, not just because sequence browser button is clicked (e.g., any module can start/stop replay on any sequence browser node at any time).

SequenceReplay/SequenceReplay.py
```diff
@@ -255,95 +256,103 @@ def setParameterNode(self, inputParameterNode: Optional[SequenceReplayParameterN
...
+
             for nodeId in transformableNodeIds:
                 node = slicer.mrmlScene.GetNodeByID(nodeId)
-                checkBox = qt.QCheckBox(node.GetName())
-                checkboxesLayout.addWidget(checkBox)
+                if node:
+                    checkBox = qt.QCheckBox(node.GetName())
+                    checkboxesLayout.addWidget(checkBox)
+
+    def _clearLayout(self, layout):
```
lassoan, line 349: you can make it a staticmethod to indicate that it is just some global utility function and does not use or modify `self`.

lassoan (commented):
Thanks for the update, I've added a couple of comments inline.

PR #225, opened 2024-11-26
fix: import i18n in Ultrasound class

mcfly3001:
An exception is raised when using the Guidelet because the new localization feature was missing an import for `_` in the Ultrasound.py. Fix by adding the respective import.


lassoan (approved):
Good catch, thank you!

